PR #15, opened 2021-01-14
Adds initial implementation of the JWT package

austingebauer:
## Description

This PR adds an initial implementation of the JWT package, which provides signature verification and claims set validation for [JWTs](https://tools.ietf.org/html/rfc7519) of the [JWS](https://tools.ietf.org/html/rfc7519) form.

The code in this package is similar to that in both the [vault-plugin-auth-jwt](https://github.com/hashicorp/vault-plugin-auth-jwt/blob/master/path_login.go#L47) and [consul](https://github.com/hashicorp/consul-enterprise/blob/master/internal/go-sso/oidcauth/jwt.go#L24) JWT auth implementations.

## Testing

I've integrated and tested the code in this PR in the [common-jwt-lib](https://github.com/hashicorp/vault-plugin-auth-jwt/compare/common-jwt-lib) branch of vault-plugin-auth-jwt.

Tests are intentionally absent from this PR in order to first agree on the API and behavior of the package. I'll be adding tests and additional documentation in a subsequent PR once there is agreement.

### DIFF
--- a/jwt/jwt.go
+++ b/jwt/jwt.go
@@ -89,7 +89,8 @@ type Expected struct {
 //  3. It's valid with respect to the current time. This means that the current
 //     time must be within the times (inclusive) given by the "nbf" (Not Before)
 //     and "exp" (Expiration Time) claims and after the time given by the "iat"
-//     (Issued At) claim, with configurable leeway.
+//     (Issued At) claim, with configurable leeway. See Expected.Now() for details
+//     on how the current time is provided for validation.
 func (v *Validator) Validate(ctx context.Context, token string, expected Expected) (map[string]interface{}, error) {
 	// First, verify the signature to ensure subsequent validation is against verified claims
 	allClaims, err := v.keySet.VerifySignature(ctx, token)
--- a/jwt/jwt.go
+++ b/jwt/jwt.go
@@ -1,0 +1,259 @@
+package jwt
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"time"
+
+	"gopkg.in/square/go-jose.v2"
+	"gopkg.in/square/go-jose.v2/jwt"
+)
+
+const (
+	defaultLeewaySeconds = 150
+)
+
+// Validator validates JSON Web Tokens (JWT) by providing signature
+// verification and claims set validation.
+type Validator struct {
+	keySet KeySet
+}
+
+// NewValidator returns a Validator that uses the given KeySet to verify JWT signatures.
+func NewValidator(keySet KeySet) (*Validator, error) {
+	if keySet == nil {
+		return nil, errors.New("keySet must not be nil")
+	}
+
+	return &Validator{
+		keySet: keySet,
+	}, nil
+}
+
+// Expected defines the expected claims values to assert when validating a JWT.
+// For claims that involve validation of the JWT with respect to time, leeway
+// fields are provided to account for potential clock skew.
+type Expected struct {
+	// The expected JWT "iss" (issuer) claim value. If empty, validation is skipped.
+	Issuer string
+
+	// The expected JWT "sub" (subject) claim value. If empty, validation is skipped.
+	Subject string
+
+	// The expected JWT "jti" (JWT ID) claim value. If empty, validation is skipped.
+	ID string
+
+	// The list of expected JWT "aud" (audience) claim values to match against.
+	// The JWT claim will be considered valid if it matches any of the expected
+	// audiences. If there are no expected audiences, then the presence of any
+	// audience in the JWT will render it invalid.
+	Audiences []string
+
+	// SigningAlgorithms provides the list of expected JWS "alg" (algorithm) header
+	// parameter values to match against. The JWS header parameter will be considered
+	// valid if it matches any of the expected signing algorithms. If empty, defaults
+	// to "RS256" as defined in https://www.rfc-editor.org/rfc/rfc7518.html#section-3.1.
+	SigningAlgorithms []string
+
+	// NotBeforeLeeway provides the option to set an amount of leeway to use when
+	// validating the "nbf" (Not Before) claim. If the duration is zero or not
+	// provided, a default leeway of 150 seconds will be used. If the duration is
+	// negative, no leeway will be used.
+	NotBeforeLeeway time.Duration
+
+	// ExpirationLeeway provides the option to set an amount of leeway to use when
+	// validating the "exp" (Expiration Time) claim. If the duration is zero or not
+	// provided, a default leeway of 150 seconds will be used. If the duration is
+	// negative, no leeway will be used.
+	ExpirationLeeway time.Duration
+
+	// ClockSkewLeeway provides the option to set an amount of leeway to use when
+	// validating the "nbf" (Not Before), "exp" (Expiration Time), and "iat" (Issued At)
+	// claims. If the duration is zero or not provided, a default leeway of 60 seconds
+	// will be used. If the duration is negative, no leeway will be used.
+	ClockSkewLeeway time.Duration
+
+	// Now provides the option to specify a func for determining what the current time is.
+	// The func will be used to provide the current time when validating a JWT with respect to
+	// the "nbf" (Not Before), "exp" (Expiration Time), and "iat" (Issued At) claims. If not
+	// provided, defaults to returning time.Now().
+	Now func() time.Time
+}
+
+// Validate validates JWTs of the JWS compact serialization form.
+//
+// The given JWT is considered valid if:
+//  1. Its signature is successfully verified.
+//  2. Its claims set and header parameter values match what's given by Expected.
+//  3. It's valid with respect to the current time. This means that the current
+//     time must be within the times (inclusive) given by the "nbf" (Not Before)
+//     and "exp" (Expiration Time) claims and after the time given by the "iat"
+//     (Issued At) claim, with configurable leeway.
+func (v *Validator) Validate(ctx context.Context, token string, expected Expected) (map[string]interface{}, error) {
+	// First, verify the signature to ensure subsequent validation is against verified claims
+	allClaims, err := v.keySet.VerifySignature(ctx, token)
+	if err != nil {
+		return nil, fmt.Errorf("error verifying token signature: %w", err)
+	}
+
+	// Validate the signing algorithm in the JWS header
+	if err := validateSigningAlgorithm(token, expected.SigningAlgorithms); err != nil {
+		return nil, fmt.Errorf("invalid algorithm (alg) header parameter: %w", err)
+	}
+
+	// Unmarshal all claims into the set of public JWT registered claims
+	claims := jwt.Claims{}
+	allClaimsJSON, err := json.Marshal(allClaims)
+	if err != nil {
+		return nil, err
+	}
+	if err := json.Unmarshal(allClaimsJSON, &claims); err != nil {
+		return nil, err
+	}
+
+	// At least one of the "nbf" (Not Before), "exp" (Expiration Time), or "iat" (Issued At)
+	// claims are required to be set.
+	if claims.IssuedAt == nil {
+		claims.IssuedAt = new(jwt.NumericDate)
+	}
+	if claims.Expiry == nil {
+		claims.Expiry = new(jwt.NumericDate)
+	}
+	if claims.NotBefore == nil {
+		claims.NotBefore = new(jwt.NumericDate)
+	}
+	if *claims.IssuedAt == 0 && *claims.Expiry == 0 && *claims.NotBefore == 0 {
+		return nil, errors.New("no issued at (iat), not before (nbf), or expiration time (exp) claims in token")
+	}
+
+	// If "exp" (Expiration Time) is not set, then set it to the latest of
+	// either the "iat" (Issued At) or "nbf" (Not Before) claims plus leeway.
+	if *claims.Expiry == 0 {
+		latestStart := *claims.IssuedAt
+		if *claims.NotBefore > *claims.IssuedAt {
+			latestStart = *claims.NotBefore
+		}
+		leeway := expected.ExpirationLeeway.Seconds()
+		if expected.ExpirationLeeway.Seconds() < 0 {
+			leeway = 0
+		} else if expected.ExpirationLeeway.Seconds() == 0 {
+			leeway = defaultLeewaySeconds
+		}
+		*claims.Expiry = jwt.NumericDate(int64(latestStart) + int64(leeway))
+	}
+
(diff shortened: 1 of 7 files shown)


jimlambrt:
Just an overall observation about errors in the package: as implemented callers are forced to use string matching to test for different errors.  perhaps, we should define a set of testable errors (via errors.Is) and always wrap them with context info before returning.  More aligned with how the oidc package does this.

jwt/keyset.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	return &OIDCDiscoveryKeySet{
+		provider: provider,
+	}, nil
+}
+
+// VerifySignature parses the given JWT, verifies its signature using discovered JWKS keys, and
+// returns the claims in its payload. The given JWT must be of the JWS compact serialization form.
+func (ks OIDCDiscoveryKeySet) VerifySignature(ctx context.Context, token string) (map[string]interface{}, error) {
```
jimlambrt, line 69: I wonder if this is needed... could we just do something like: https://github.com/hashicorp/cap/tree/jimlambrt-jwt-idea

It would leverage the JSONWebKeySet type you have to have... and I don't see any need for the other capabilities provided by the oidc.Provider

jwt/keyset.go
```diff
@@ -0,0 +1,217 @@
...
+		return nil, err
+	}
+
+	return allClaims, nil
+}
+
+// NewJSONWebKeySet returns a KeySet that verifies JWT signatures using keys from the JSON Web
+// Key Set (JWKS) at the given jwksURL. The client used to obtain the remote JWKS will verify
+// server certificates using the root certificates provided by jwksCAPEM.
+func NewJSONWebKeySet(ctx context.Context, jwksURL string, jwksCAPEM string) (KeySet, error) {
```
jimlambrt, line 94: same question as above about the certs...

jwt/keyset.go
```diff
@@ -0,0 +1,217 @@
...
+type OIDCDiscoveryKeySet struct {
+	provider *oidc.Provider
+}
+
+// JSONWebKeySet verifies JWT signatures using keys obtained from a JWKS URL.
+type JSONWebKeySet struct {
+	remoteJWKS oidc.KeySet
+}
+
+// StaticKeySet verifies JWT signatures using local PEM-encoded public keys.
```
jimlambrt, line 38: Isn't just "public keys"?   Also, why not use the []crypto.PublicKey vs []interface{}?

jwt/keyset.go
```diff
@@ -0,0 +1,217 @@
...
+	if err := json.Unmarshal(payload, &allClaims); err != nil {
+		return nil, err
+	}
+
+	return allClaims, nil
+}
+
+// NewStaticKeySet returns a KeySet that verifies JWT signatures using PEM-encoded public keys.
+// The given publicKeys must be of PEM-encoded x509 certificate or PKIX public key forms.
+func NewStaticKeySet(publicKeys []string) (KeySet, error) {
```
jimlambrt, line 128: could we just accept []crypto.PublicKey directly and provide helper functions to parse PEMs and convert Certs to PublicKeys?  Is there an advantage to accepting PEMs?

jimlambrt (changes requested):
Looks good... just a few suggestions and I'm happy to discuss in real-time.

jwt/jwt.go
```diff
@@ -0,0 +1,259 @@
...
+	// The expected JWT "iss" (issuer) claim value. If empty, validation is skipped.
+	Issuer string
+
+	// The expected JWT "sub" (subject) claim value. If empty, validation is skipped.
+	Subject string
+
+	// The expected JWT "jti" (JWT ID) claim value. If empty, validation is skipped.
+	ID string
+
+	// The list of expected JWT "aud" (audience) claim values to match against.
```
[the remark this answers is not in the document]
austingebauer: I originally included this to maintain a behavior in the Vault plugin. I like the idea of consistency in the library wrt validation, so I've changed to skip on empty in 825205c. The [behavior](https://github.com/hashicorp/vault-plugin-auth-jwt/blob/master/path_login.go#L181) will be maintained outside of the library in the Vault plugin instead.

jwt/jwt.go
```diff
@@ -0,0 +1,259 @@
...
+	// provided, defaults to returning time.Now().
+	Now func() time.Time
+}
+
+// Validate validates JWTs of the JWS compact serialization form.
+//
+// The given JWT is considered valid if:
+//  1. Its signature is successfully verified.
+//  2. Its claims set and header parameter values match what's given by Expected.
+//  3. It's valid with respect to the current time. This means that the current
```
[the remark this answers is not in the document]
austingebauer: Good idea. Done in c0a761c.

jwt/jwt.go
```diff
@@ -0,0 +1,259 @@
...
+	// The expected JWT "jti" (JWT ID) claim value. If empty, validation is skipped.
+	ID string
+
+	// The list of expected JWT "aud" (audience) claim values to match against.
+	// The JWT claim will be considered valid if it matches any of the expected
+	// audiences. If there are no expected audiences, then the presence of any
+	// audience in the JWT will render it invalid.
+	Audiences []string
+
+	// SigningAlgorithms provides the list of expected JWS "alg" (algorithm) header
```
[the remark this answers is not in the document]
austingebauer: Discussed moving [oidc/algs.go](https://github.com/hashicorp/cap/blob/main/oidc/algs.go) to a common package and restricting `Expected.SigningAlgorithms` to the set of algorithms defined there. I'm happy to make that change if there is agreement on that.

jwt/jwt.go
```diff
@@ -0,0 +1,259 @@
...
+		return nil, errors.New("invalid issued at (iat) claim: token issued in the future")
+	}
+
+	return allClaims, nil
+}
+
+// validateSigningAlgorithm checks whether the JWS "alg" (Algorithm) header
+// parameter value for the given JWT matches any given in expectedAlgorithms.
+// If expectedAlgorithms is empty, "RS256" will be expected by default.
+func validateSigningAlgorithm(token string, expectedAlgorithms []string) error {
```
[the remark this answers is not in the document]
austingebauer: Will address along with comment on `Expected.SigningAlgorithms`.

jwt/keyset.go
```diff
@@ -0,0 +1,217 @@
...
+// StaticKeySet verifies JWT signatures using local PEM-encoded public keys.
+type StaticKeySet struct {
+	publicKeys []interface{}
+}
+
+// NewOIDCDiscoveryKeySet returns a KeySet that verifies JWT signatures using keys from the
+// JSON Web Key Set (JWKS) published in the discovery document at the given discoveryURL.
+// The client used to obtain the remote keys will verify server certificates using the root
+// certificates provided by discoveryCAPEM.
+func NewOIDCDiscoveryKeySet(ctx context.Context, discoveryURL string, discoveryCAPEM string) (KeySet, error) {
```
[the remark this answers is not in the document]
austingebauer: The certs are optional in both `NewOIDCDiscoveryKeySet` and `NewJSONWebKeySet`. I've added a comment to clarify that in c9303e2.

jwt/keyset.go
austingebauer: The certs are optional in both NewOIDCDiscoveryKeySet and NewJSONWebKeySet. I've added a comment to clarify that in c9303e2.

jwt/keyset.go
austingebauer: I like this idea. I don't think we need to hold on to `oidc.Provider` when all it really does is retrieve the `jwks_uri` from the OIDC discovery document and create a remote key set for verification. This would also make the verification more consistent with `JSONWebKeySet`. 

Thanks for putting together the example code!

jwt/keyset.go
austingebauer: Yep, I plan to change this to just "public keys" as we discussed. I originally used PEMs because that's the way it was in the Vault plugin. I think it makes more sense to accept `[]crypto.PublicKey` directly now that this is in a library. I also like the idea of providing a helper function to parse PEMs.

jwt/keyset.go
austingebauer: Will address along with comment on `StaticKeySet`.

jwt/keyset.go
austingebauer: I've changed this to accept `[]crypto.PublicKey` in d944a7e. Let me know your thoughts.

jwt/keyset.go
austingebauer: Addressed in d944a7e. Let me know your thoughts.

jwt/keyset.go
jimlambrt: I like it.  ty!

jwt/jwt.go
```diff
@@ -0,0 +1,256 @@
...
+	for _, v := range expectedAudiences {
+		if contains(audClaim, v) {
+			return nil
+		}
+	}
+
+	return errors.New("audience claim does not match any expected audience")
+}
+
+func contains(sl []string, st string) bool {
```
calvn, line 249: This looks exactly like `StrListContains` from `oidc/internal/strutils`. @jimlambrt thoughts on moving `oidc/internal` into the top-level `cap` directory?

jwt/keyset.go
```diff
@@ -0,0 +1,211 @@
...
+		return ctx, nil
+	}
+
+	certPool := x509.NewCertPool()
+	if ok := certPool.AppendCertsFromPEM([]byte(caPEM)); !ok {
+		return nil, errors.New("could not parse CA PEM value successfully")
+	}
+
+	tr := cleanhttp.DefaultPooledTransport()
+	if certPool != nil {
```
calvn, line 199: I don't think `certPool` can ever be `nil` if it is instantiated by `x509.NewCertPool()`.

jwt/jwt.go
```diff
@@ -0,0 +1,259 @@
...
+	"errors"
+	"fmt"
+	"time"
+
+	"gopkg.in/square/go-jose.v2"
+	"gopkg.in/square/go-jose.v2/jwt"
+)
+
+const (
+	defaultLeewaySeconds = 150
```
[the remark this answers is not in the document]
austingebauer: I agree. I've exported and documented it in e6e4b2d.

jwt/keyset.go
austingebauer: That's true. Removed in 5d29657.

jwt/keyset.go
austingebauer: I've removed the need for using an `oidc.Provider` for obtaining keys via OIDC discovery in abb14cd. The code is very similar to that in your example branch with some minor adjustments and comments. Let me know your thoughts 👍

jwt/jwt.go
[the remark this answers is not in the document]
austingebauer: I've made changes to restrict the set of expected `SigningAlgorithms` in d8a696d. I decided to copy the constants from [oidc/algs.go](https://github.com/hashicorp/cap/blob/main/oidc/algs.go) into the jwt package instead of creating a "common" or "shared" package. This approach seemed okay for constants. I'm happy to consider alternative ideas around adding them to a separate package though.

jwt/jwt.go
jimlambrt: Hum... it's such a small simple bit of behavior, I'm okay with it being copied.

jwt/keyset.go
```diff
@@ -0,0 +1,239 @@
...
+	remoteJWKS oidc.KeySet
+}
+
+// staticKeySet verifies JWT signatures using local public keys.
+type staticKeySet struct {
+	publicKeys []crypto.PublicKey
+}
+
+// NewOIDCDiscoveryKeySet returns a KeySet that verifies JWT signatures using keys from the
+// JSON Web Key Set (JWKS) published in the discovery document at the given discoveryURL.
```
jimlambrt, line 45: Hum... if we're going to use discoveryURL as a name, then we should specify that ".well-known/openid-configuration" shouldn't be included.  We could also change the name to "issuer", which explicity doesn't include the URL fragment.

jimlambrt (changes requested):
Looks great. One very small suggestion/nit.

jwt/keyset.go
austingebauer: I agree. I think changing to "issuer" makes more sense here. Done in decf7c6.

jimlambrt (approved):
Looks great.  ty!

jwt/jwt.go
```diff
@@ -0,0 +1,265 @@
...
+
+// validateSigningAlgorithm checks whether the JWS "alg" (Algorithm) header
+// parameter value for the given JWT matches any given in expectedAlgorithms.
+// If expectedAlgorithms is empty, RS256 will be expected by default.
+func validateSigningAlgorithm(token string, expectedAlgorithms []Alg) error {
+	for _, expected := range expectedAlgorithms {
+		if !supportedAlgorithms[expected] {
+			return fmt.Errorf("unsupported signing algorithm %q", expected)
+		}
+	}
```
kalafut, line 212: Is this check something we'll be able to run at config time too?

jwt/keyset.go
```diff
@@ -0,0 +1,239 @@
...
+
+	// Unmarshal the response body to obtain the issuer and JWKS URL
+	type providerJSON struct {
+		Issuer  string `json:"issuer"`
+		JWKSURL string `json:"jwks_uri"`
+	}
+	var p providerJSON
+	err = unmarshalResp(resp, body, &p)
+	if err != nil {
+		return nil, fmt.Errorf("failed to decode OIDC discovery document: %v", err)
```
kalafut, line 93:
```suggestion
		return nil, fmt.Errorf("failed to decode OIDC discovery document: %w", err)
```

jwt/keyset.go
```diff
@@ -0,0 +1,239 @@
...
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("unable to read response body: %v", err)
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("%s: %s", resp.Status, body)
+	}
+
+	// Unmarshal the response body to obtain the issuer and JWKS URL
+	type providerJSON struct {
```
kalafut, line 86: A type seems superfluous here. Maybe just:
```
	var p struct {
		Issuer  string `json:"issuer"`
		JWKSURL string `json:"jwks_uri"`
	}
```

austingebauer: Ah, I forgot about the anonymous struct. Thanks! Done in 0792e44.

calvn (approved):

jwt/jwt.go
austingebauer: I've introduced a function in 72a2a3b that'll allow us to run this check at config time.

PR #19, opened 2021-02-07
support for the WithNonce(...) oidc.Request option

jimlambrt:
Support for the WithNonce(...) option for NewRequest(..) allows devs to specify their own nonce value.  

This new option is required, if the developer wishes to encode all required Request info into a Request's State.  Encoding all a Request's info into its State would remove the requirement of a Request cache/storage from the developer's application.  We already support a WithState(...) option, and the Request's Nonce was the only Request field which the developer could not specify or change.


austingebauer (approved):
👍

PR #20, opened 2021-02-11
Add support for getting a hash of a Config, so you can compare them for equality

jimlambrt:
This feature allows you to compare cached providers for configuration equality.


oidc/config.go
```diff
@@ -114,6 +119,60 @@ func NewConfig(issuer string, clientID string, clientSecret ClientSecret, suppor
...
+	var h uint64
+	var err error
+	args := []string{}
+	args = append(args, c.Issuer, c.ClientID, string(c.ClientSecret), c.ProviderCA)
+	for _, a := range c.SupportedSigningAlgs {
+		args = append(args, string(a))
+	}
+	args = append(args, c.Scopes...)
+	args = append(args, c.Audiences...)
+	args = append(args, c.AllowedRedirectURLs...)
```
mgaffney, line 134: Should the slices be sorted before hashing?

jimlambrt: Yes.  Ty!

mgaffney (approved):

oidc/config.go
```diff
@@ -114,6 +120,78 @@ func NewConfig(issuer string, clientID string, clientSecret ClientSecret, suppor
...
+
+	redirects := make([]string, 0, len(c.AllowedRedirectURLs))
+	redirects = append(redirects, c.AllowedRedirectURLs...)
+
+	sort.Strings(algs)
+	sort.Strings(scopes)
+	sort.Strings(audiences)
+	sort.Strings(redirects)
+
+	args := make([]string, 0, len(algs)+len(scopes)+len(audiences)+len(redirects)+4)
```
austingebauer, line 148: I think the final append of the `NowFunc` value would make the capacity of `args` need a `+5` instead of `+4`? No functional difference, but I think the final append would end up allocating a new underlying array since it'd be over capacity.

austingebauer (approved):
LGTM

oidc/config.go
jimlambrt: ty! great catch.  I've rearrage things a bit to make it easier to align the capcity with the number of args.

PR #25, opened 2021-03-24
Port OpenURL and IsWSL from Vault's JWT auth repo

jefferai:
These are useful for Boundary too so putting them in the shared repo
seems like the right move. There aren't unit tests for these as they're
very system specific but they're easy to test.

The logic hasn't changed in terms of OS-specific behavior, however,
instead of unilaterally printing to stderr if there is a problem reading
/proc/version, it will construct a multierror instead and return that
along with the error (if any) from the executed command. This is a
cleaner way to deal with an issue here as it lets the calling command
decide how to represent the error to the user.


jimlambrt (approved):
ty!

austingebauer (approved):
👍

PR #26, opened 2021-03-24
Set specific version of golang.org/x/oauth2

austingebauer:
## Description

This PR resets the version requirement for the `golang.org/x/oauth2` dependency, which was set to a specific version in PR  https://github.com/hashicorp/cap/pull/14. I missed seeing that this was changed when reviewing PR https://github.com/hashicorp/cap/pull/25. See PR https://github.com/hashicorp/cap/pull/14 for more context on why we're setting this specific version. 

An alternative would be to fix the etcd/grpc incompatibility in Vault based on guidance from https://github.com/etcd-io/etcd/issues/12124.

## Similar PRs

- https://github.com/hashicorp/vault-plugin-auth-azure/pull/43


jimlambrt (approved):
I missed the change as well. Great catch!

PR #27, opened 2021-03-26
add discovery info for: supported subject types and response types plus unit tests for the discovery endpoint

jimlambrt:
…


austingebauer (approved):
👍

PR #29, opened 2021-04-03
OIDC: add the ability to run a TestProvider without TLS and start a TestProvider with using a type that implements the TestingT interface

jimlambrt:
These changes would allow devs to run the TestProvider outside of "tests" if they need a very simple Provider. 

I've also included a NewTestingLogger which implements the TestingT interface and logs errors via hclog

@jefferai, @austingebauer : I just pushed a couple commits to this PR which add the ability to present a login form for interactive testing.


oidc/examples/cli/main.go
```diff
@@ -71,22 +74,86 @@ func main() {
...
+	var env map[string]interface{}
+	var tp *oidc.TestProvider
+	if *useTestProvider {
+		l, err := oidc.NewTestingLogger(hclog.New(nil))
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "%s\n\n", err)
+			return
+		}
+		// Generate a key to sign JWTs with throughout most test cases
+		priv, err := rsa.GenerateKey(rand.Reader, 2048)
```
jefferai, line 104: I don't feel strongly about this, but you may want to not use RSA2048. I realize it's always available because it's required by the standard, but it's also not super desirable. So given this is an example we may want to encourage more secure options.

austingebauer (approved):

PR #36, opened 2021-05-17
update coreos go-odic dependency to V3 resolve JWKs caching issue.

jimlambrt:
This PR for the go-oidc dependency resolves a caching issue for JWKs: https://github.com/coreos/go-oidc/pull/259

I've also taken the opportunity to update all other dependencies to the latest versions except for golang.org/x/oauth2.  It appears the etcd team is getting close to release v3.5.0 which would allow us to update that dependency as well in the near future.


austingebauer:
I think we'll also need to update the `jwt` package. I can take care of that.

jimlambrt:
@austingebauer I thought, I got all the `jwt` package changes required. I believe you only reference the coreos package in ` jwt/keyset.go`

austingebauer (approved):
You're correct. I hadn't pulled down the V3 source and thought there was a breaking change. 

Everything looks good. I tested this out on the Vault plugin as well. Thanks!

PR #44, opened 2021-12-16
chore: Upgrade dependencies

jimlambrt:
Notice the upgraded Oauth dep.


austingebauer:
Thanks! Testing this out on the Vault side. Will get back shortly and approve if all is good.

austingebauer (approved):
Tested this in Vault. Works as expected. Thanks, @jimlambrt!

PR #45, opened 2022-01-07
feature (ldap): Add support for LDAP/AD authentication

jimlambrt:


jimlambrt:
There are a few more tests to write, but I believe the required capabilities are ready for review.

jimlambrt:
I just discovered that a pkg I'm using for the TestDirectory has a GPL license, so I need to figure out the best path forward to eliminate that dependency.

PR #48, opened 2022-02-09
Validate sub and aud are present in ID token

tbalasavage:
This PR adds additional validation in order ensure that the `sub` and `aud` claims of an ID token are populated.


jimlambrt (approved):
ty!

PR #50, opened 2022-05-02
Add option to allow all of IAT, NBF, and EXP to be missing

swenson:
This is necessary for Kubernetes 1.20 auth tokens, which don't include any of `iat`/`exp`/`nbf`, so this library has no way to deal with them. (Which blocks `vault-plugin-auth-kubernetes` from using this library, since we still support Kubernetes 1.20.)

Also did a small amount of cleanup:
* Generate a single 4096-bit RSA key in the tests, which speeds them up by a few seconds.
* Update go.mod in ldap cli example to use commit that exists (otherwise the tests appear to fail)


jwt/jwt.go
```diff
@@ -125,8 +157,14 @@ func (v *Validator) Validate(ctx context.Context, token string, expected Expecte
 	if claims.NotBefore == nil {
 		claims.NotBefore = new(jwt.NumericDate)
 	}
+	// At least one of the "nbf" (Not Before), "exp" (Expiration Time), or "iat" (Issued At)
+	// claims are required to be set unless allowMissingIatExpNbf is set.
 	if *claims.IssuedAt == 0 && *claims.Expiry == 0 && *claims.NotBefore == 0 {
-		return nil, errors.New("no issued at (iat), not before (nbf), or expiration time (exp) claims in token")
+		if allowMissingIatExpNbf {
+			return allClaims, nil
+		} else {
```
austingebauer, line 165: nit: can drop the else here

jwt/jwt_test.go
```diff
@@ -281,6 +290,61 @@ func TestValidator_Validate_Valid_JWT(t *testing.T) {
 	}
 }
 
+func TestValidator_NoExpIatNbf(t *testing.T) {
```
austingebauer, line 293: Thanks for adding this!

austingebauer (approved):
This looks good to me! Thanks, @swenson.

jwt/jwt_test.go
swenson: 👍

jwt/jwt.go
swenson: 👍

swenson:
Thanks!

PR #57, opened 2022-11-29
feature (odic): Add support for ProviderConfig 

### LINKED ISSUE #55: Add support for static OIDC discovery metadata and JWKs
**Is your feature request related to a problem? Please describe.**

This library supports OIDC implicit flow (`response_type=id_token`) which doesn't need direct communication between RP and OP, but it needs direct communication for getting OIDC discovery metadata and JWKs currently.

As a specific use case, I'd like to use https://github.com/hashicorp/vault-plugin-auth-jwt to support authentication by OIDC without direct communication. Currently, CAP does not have the API, so need to implement this using a lot of reflect package as follows, which does not seem like a very good way to do it.

https://github.com/openstandia/vault-plugin-auth-jwt/commit/9eb0e93aed0385f0044e904077ac51c933fba7ae

**Describe the solution you'd like**
For OIDC implicit flow without direct communication, I'd like to add support for static OIDC discovery metadata and JWKs.
Currently, `provider.go` has a factory func. I propose adding feature which creates the provider from static json documents for OIDC discovery metadata and JWKs.

https://github.com/hashicorp/cap/blob/bb2447f2023e5160509bd6bdf25db8cc9ab8e6fd/oidc/provider.go#L60

jimlambrt:
`ProviderConfig` which creates a provider that doesn't support OIDC discovery. It's probably better to use NewProvider(...) with discovery whenever possible

Fixes #55


austingebauer (approved):
Looks great! Thanks for doing this @jimlambrt 👍

PR #58, opened 2022-12-16
feature (ldap): add WithUserAttributes option

jimlambrt:
The new WithUserAttributes option allows callers to request the attributes be returned for the authenticating user.


austingebauer (approved):
LGTM! I'm thinking we could eventually use this in Vault for alias and token metadata population. Thanks @jimlambrt.

PR #62, opened 2023-02-03
Replace square/go-jose with go-jose/go-jose/v3

benashz:
The square/go-jose project is deprecated, and is set to be archived on Feb. 27th, 2023. Luckily the project has been moved to go-jose/go-jose.

ref: https://github.com/square/go-jose#go-jose-v1v2-deprecated


austingebauer (approved):
Thanks, @benashz! LGTM once we can see the tests passing.

PR #63, opened 2023-02-03
CI: ensure that all build/test failures are properly detected

benashz:
Previously, no build/test failures would result in a non-zero exit status. This PR should now fail the build/tests steps whenever an error is encountered.

Here's an example of the issue:
https://github.com/hashicorp/cap/actions/runs/3989282340/jobs/6841516383#step:6:10

Other fixes/updates:
- enable verbose testing. (open to dropping this, if it is not desired)
- fix panic in `TestClient_connect`
- bump the min go version to 1.19
- drop support for dep

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -16,7 +16,7 @@ jobs:
     - name: Set up Go 1.x
       uses: actions/setup-go@v2
       with:
-        go-version: ^1.13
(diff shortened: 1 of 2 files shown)


austingebauer (approved):
I just noticed the test failure wasn't reported from the other PR then saw this 👍 Thanks for doing this, @benashz!

PR #75, opened 2023-06-06
feat (oidc): better error for max age

jimlambrt:


oidc/provider.go
```diff
@@ -556,7 +556,7 @@ func (p *Provider) VerifyIDToken(ctx context.Context, t IDToken, oidcRequest Req
 		}
 		authTime := time.Unix(int64(atClaim), 0)
 		if !authTime.Add(leeway).After(authAfter) {
-			return nil, fmt.Errorf("%s: auth_time (%s) is beyond max age (%d): %w", op, authTime, secs, ErrExpiredAuthTime)
+			return nil, fmt.Errorf("%s: issued at (%s) is beyond max age (%d / %s): %w", op, authTime, secs, authAfter, ErrExpiredAuthTime)
```
austingebauer, line 559: Nit: I think keeping `auth_time` is more precise since it can be different from the issued at (`iat`) claim.

austingebauer (approved):
Thanks!

PR #80, opened 2023-08-03
Add tests for Config and ServiceProvider

hcjulz:
This PR adds table tests for creating new configs and cleans up a little bit of code.

### DIFF
--- a/saml/config.go
+++ b/saml/config.go
@@ -1,105 +1,73 @@
 package saml
 
 import (
-	"crypto/rsa"
-	"crypto/tls"
 	"errors"
 	"fmt"
 	"net/url"
 	"time"
 
 	"github.com/hashicorp/cap/oidc"
-	"github.com/hashicorp/cap/saml/models/core"
 )
 
 var ErrInvalidTLSCert = errors.New("invalid tls certificate")
 
 type ValidUntilFunc func() time.Time
 
-type IDPConfig struct {
-	// SSOServiceURL is the SSO service endpoint where the authentication request
-	// should be sent to. The endpoint must match the location for the given
-	// service binding.
-	SSOServiceURL string
-
-	// TODO: Add more fields for the cert, crypto algorithm, etc.
-	// SLOServiceURL string
-}
-
 type Config struct {
-	// ServiceBinding defines the mechanism that should be used
-	// when exchanging messages between the requester (SP) and responder (IDP).
-	ServiceBinding core.ServiceBinding
-
-	// NameIDFormat controls how the users at the IDP are mapped to
-	// users at the SP during SSO.
-	NameIDFormat core.NameIDFormat
-
 	// AssertionConsumerServiceURL defines the endpoint at the SP where the IDP
-	// will redirect to with its authentication response.
+	// will redirect to with its authentication response. (required)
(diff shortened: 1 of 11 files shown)


saml/config.go
```diff
@@ -1,105 +1,73 @@
...
-	SSOServiceURL string
-
-	// TODO: Add more fields for the cert, crypto algorithm, etc.
-	// SLOServiceURL string
-}
-
 type Config struct {
-	// ServiceBinding defines the mechanism that should be used
-	// when exchanging messages between the requester (SP) and responder (IDP).
-	ServiceBinding core.ServiceBinding
```
hcjulz, line 32: We don't really need this here. The idea in the beginning was to configure what kind of binding should be used, but that can be decided when creating the SAML AuthnRequest.

saml/config.go
```diff
@@ -1,105 +1,73 @@
...
-
-	// NameIDFormat controls how the users at the IDP are mapped to
-	// users at the SP during SSO.
-	NameIDFormat core.NameIDFormat
```
hcjulz, line 36: I made this the default for the AuthnRequest and the plan is to add options to the create AuthnRequest methods later.

saml/config.go
```diff
@@ -1,105 +1,73 @@
...
 	MetadataURL *url.URL
 
-	// ValidUntil is a function that defines until the generate service provider metadata
-	// document is valid.
+	// ValidUntil is a function that defines until the generated service provider metadata
+	// document is valid. (optional)
 	ValidUntil ValidUntilFunc
-
-	// Certificate is used to sign SAML Authentication Requests.
-	Certificate *tls.Certificate
```
hcjulz, line 56: As we don't plan to sign Auth request in first place I removed it for now. We can add it back when implementing that feature.

saml/config.go
```diff
@@ -1,105 +1,73 @@
...
-func NewConfig(entityID, acs, issuer, metadata *url.URL) *Config {
-	return &Config{
+func NewConfig(entityID, acs, issuer, metadata *url.URL) (*Config, error) {
+	const op = "saml.NewConfig"
+
+	cfg := &Config{
 		EntityID:                    entityID,
 		Issuer:                      issuer,
 		AssertionConsumerServiceURL: acs,
 		ValidUntil:                  DefaultValidUntil,
```
hcjulz, line 42: I left this for now. I'd make this an option later too.

saml/config.go
austingebauer: Sounds good to me 👍

saml/config.go
```diff
@@ -1,105 +1,73 @@
...
-	// when exchanging messages between the requester (SP) and responder (IDP).
-	ServiceBinding core.ServiceBinding
-
-	// NameIDFormat controls how the users at the IDP are mapped to
-	// users at the SP during SSO.
-	NameIDFormat core.NameIDFormat
-
 	// AssertionConsumerServiceURL defines the endpoint at the SP where the IDP
-	// will redirect to with its authentication response.
+	// will redirect to with its authentication response. (required)
```
austingebauer, line 18: Nice addition of required/optional for the godocs.

austingebauer:
A little late here. Thanks, @hcjulz. Looks good ✅

PR #84, opened 2023-08-11
Adds ability to provide IdP metadata as XML or individual parameters

austingebauer:
## Overview

This PR adds the ability to specify identity provider metadata as either an XML document or individual parameters. This is a common way that identity providers offer metadata to configure SAML service providers. I've introduced them as functional options to the `Config` struct.

Additionally, I've changed the `Config` struct to take in strings instead of `url.URL` for URL parameters. This felt more user friendly and aligns with my general experience using libraries that take in URLs (including `cap/oidc` and `cap/jwt`). I'm happy to change this back if there is disagreement.

## Testing

I've manually tested both XML identity provider metadata and individual parameters using Okta and the demo program. If everything looks reasonable, I'll add testing to this PR.

### DIFF
--- a/saml/config.go
+++ b/saml/config.go
@@ -67,14 +67,14 @@ func (c *MetadataParameters) Validate() error {
 		return fmt.Errorf("issuer not set")
 	}
 	if _, err := url.Parse(c.Issuer); err != nil {
-		return fmt.Errorf("failed to parse Issuer: %w", err)
+		return fmt.Errorf("provided Issuer is not a valid URL: %w", err)
 	}
 
 	if c.SingleSignOnURL == "" {
 		return fmt.Errorf("SSO URL not set")
 	}
 	if _, err := url.Parse(c.SingleSignOnURL); err != nil {
-		return fmt.Errorf("failed to parse SSO URL: %w", err)
+		return fmt.Errorf("provided SSO URL is not a valid URL: %w", err)
 	}
 
 	if _, err := parsePEMCertificate([]byte(c.IDPCertificate)); err != nil {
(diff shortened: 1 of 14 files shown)


saml/config.go
```diff
@@ -60,55 +158,97 @@ func NewConfig(entityID, acs, issuer, metadata *url.URL) (*Config, error) {
...
+		return fmt.Errorf("%s: failed to parse ACS URL: %w", op, ErrInvalidParameter)
 	}
 
-	if c.Issuer == nil {
-		return fmt.Errorf("%s: Issuer not set: %w", op, oidc.ErrInvalidParameter)
+	if c.EntityID == "" {
+		return fmt.Errorf("%s: EntityID not set: %w", op, ErrInvalidParameter)
+	}
+	if _, err := url.Parse(c.EntityID); err != nil {
+		return fmt.Errorf("%s: failed to parse EntityID: %w", op, ErrInvalidParameter)
```
hcjulz, line 176: [non-blocking] I think providing the config params as string makes the UX better. I like that you added parsing the URLs as validation checks here 🚀 
I'm wondering however, if we should change the error message to tell the user that the provided EntityID isn't a valid URL? 🤔

saml/config.go
```diff
@@ -60,55 +158,97 @@ func NewConfig(entityID, acs, issuer, metadata *url.URL) (*Config, error) {
...
+	if c.MetadataParameters != nil {
+		if err := c.MetadataParameters.Validate(); err != nil {
+			return fmt.Errorf("%s: %s: %w", op, err.Error(), ErrInvalidParameter)
+		}
 	}
 
 	if c.GenerateAuthRequestID == nil {
 		return fmt.Errorf(
-			"%s: GenerateAuthRequestID func not provided: %w",
+			"%s: DefaultGenerateAuthRequestID func not provided: %w",
```
hcjulz, line 202: [non-blocking] Should this be just `GenerateAuthRequestID`?

hcjulz (approved):
👏 Looks good! Left two non-blocking comments. 
Thx for this!

saml/config.go
austingebauer: Whoops, done in https://github.com/hashicorp/cap/pull/84/commits/8b5e66cad3a49bf34478a922d1a1cf5945ab178e 👍

saml/config.go
austingebauer: Ah yes, that would be helpful 👍 Changing this now.

austingebauer: Done in https://github.com/hashicorp/cap/pull/84/commits/3d14e7adb1b29d6b44f764295d4b33d3a6864d31.

austingebauer:
Thanks @hcjulz! I'll merge this to keep things moving forward. I'll add more tests in a subsequent PR 👍

PR #87, opened 2023-08-24
feat (ldap): add support for search paging and dereferencing aliases

jimlambrt:


austingebauer (approved):
This looks good to me! Would be good to give it a quick test in @raymonstah branch on the Vault side.

ldap/client.go
```diff
@@ -525,15 +528,32 @@ func (c *Client) filterGroupsSearch(userDN string, username string) ([]*ldap.Ent
...
-		Attributes: []string{
-			c.conf.GroupAttr,
-		},
-		SizeLimit: math.MaxInt32,
-	})
+	var result *ldap.SearchResult
+	switch {
+	case c.conf.MaximumPageSize > 0:
+
+		result, err = c.conn.SearchWithPaging(&ldap.SearchRequest{
```
raymonstah, line 535: Can we extract the `ldap.SearchRequest` parameter out and reuse it across both `SearchWithPaging` and `Search`? I believe they are the same.

raymonstah (approved):
Looks good! Thanks for adding this 🎉

ldap/client.go
jimlambrt: ty!

PR #89, opened 2023-08-25
Fix validUntil attribute

remilapeyre:
`DescriptorCommon.ValidUntil` is a `time.Time` struct so the `omitempty` tag has no effect. This patch changes it to be a pointer so that it can be properly omitted.

Also in CreateMetadata() we set both `ValidUntil` on the `EntityDescriptorSPSSO` and the inner `SPSSODescriptor`. The spec says

	validUntil - Optional attribute indicates the expiration time of the
	metadata contained in the element and any contained elements.

so this is actually redundant.


hcjulz (approved):
🚀

austingebauer (approved):
👍

PR #90, opened 2023-08-25
Remove NameIDFormat default

remilapeyre:
NameIDFormat is

	Zero or more elements of type anyURI that enumerate the name identifier
	formats supported by this system entity acting in this role. See Section
	8.3 of [SAMLCore] for some possible values for this element.

Vault (and I think Boundary also) do not need it to be an email, the name is used as an opaque string.

I suppose most public IdP will have an email for their user but private ones may not have one and use an employee ID or a username instead.

I think it's best to keep the default configuration empty to be compatible with such systems.

I also updated some subtests that where failing the parent test when an error occured. They now do fail the subtest itself instead.

### DIFF
--- a/saml/authn_request_test.go
+++ b/saml/authn_request_test.go
@@ -68,7 +68,8 @@ func Test_CreateAuthnRequest(t *testing.T) {
 	}
 
 	for _, c := range cases {
-		t.Run(c.name, func(_ *testing.T) {
+		t.Run(c.name, func(t *testing.T) {
+			r := require.New(t)
 			got, err := provider.CreateAuthnRequest(c.id, c.binding)
 			if c.err != "" {
 				r.Error(err)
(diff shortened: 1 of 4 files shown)


hcjulz (approved):
Thx @remilapeyre 👏 LGTM
There is only one thing we might want to change as part of this PR. We have the option `WithAdditionalNameIDFormat`. Now that we don't have a default it probably makes sense to just remove this option or to change it to `WithNameIDFormat`. Wdyt?

austingebauer (approved):

remilapeyre:
> 

WithAdditionalNameIDFormat() does an `append()` to the list so it still adds something even though now the list is empty at first.

```
saml.WithAdditionalNameIDFormat(core.NameIDFormatEmail)
saml.WithAdditionalNameIDFormat(core.NameIDFormatX509SubjectName)
```

gives `[]core.NameIDFormat{core.NameIDFormatEmail, core.NameIDFormatX509SubjectName}` but I would expect


```
saml.WithNameIDFormat(core.NameIDFormatEmail)
saml.WithNameIDFormat(core.NameIDFormatX509SubjectName)
```

to give `[]core.NameIDFormat{core.NameIDFormatX509SubjectName}` so the name `WithAdditionalNameIDFormat` still looks fine to me.

PR #91, opened 2023-08-25
Set AllowCreate default value to true

remilapeyre:
The SAML2Core spec says for AllowCreate:

	A Boolean value used to indicate whether the requester grants to the identity provider, in the
	course of fulfilling the request, permission to create a new identifier or to associate an existing
	identifier representing the principal with the relying party. Defaults to "false"if not present or the entire
	element is omitted.

	[...]

	Requesters that do not make specific use of this attribute SHOULD generally set it to “true” to
	maximize interoperability.

### DIFF
--- a/saml/authn_request.go
+++ b/saml/authn_request.go
@@ -30,7 +30,7 @@ type authnRequestOptions struct {
 
 func authnRequestOptionsDefault() authnRequestOptions {
 	return authnRequestOptions{
-		allowCreate:     false,
+		allowCreate:     true,
 		nameIDFormat:    core.NameIDFormat(""),
 		forceAuthn:      false,
 		protocolBinding: core.ServiceBindingHTTPPost,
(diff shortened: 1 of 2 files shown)


saml/authn_request.go
```diff
@@ -45,10 +45,10 @@ func getAuthnRequestOptions(opt ...Option) authnRequestOptions {
 
 // AllowCreate is a Boolean value used to indicate whether the identity provider is allowed, in the course
 // of fulfilling the request, to create a new identifier to represent the principal.
-func AllowCreate() Option {
+func AllowCreate(allow bool) Option {
```
hcjulz, line 48: Given this is a flag, we might come around it without a bool param. Something linke `DisableAllowCreate()`.

hcjulz (commented):
@remilapeyre thx for the PR. 
In general this looks good to me, but I went through the specs again: [OASISCoreSpec](https://www.oasis-open.org/committees/download.php/56776/sstc-saml-core-errata-2.0-wd-07.pdf) and [SAML2Int](https://kantarainitiative.github.io/SAMLprofiles/saml2int.html#_web_browser_sso). 

In SAML2Int they say:
```
[SDP-SP04]
The <samlp:AuthnRequest> message MUST either omit the <samlp:NameIDPolicy> element (RECOMMENDED), or the element MUST contain an AllowCreate attribute of "true" and MUST NOT contain a Format attribute.
```

In OASIDCore they say about the `NameIDPolicy` element:

```
[3.4.1 Element <AuthnRequest>]
... If omitted, then any type of identifier supported by the identity provider for the requested subject can be used, constrained by any relevant deployment-specific policies, with respect to privacy, for example.
```
Then on the `AllowCreate` option they say it should generally be set to true if a `NameIDPolicy` element is set.

So this means, if no NameIDFormat element is set, the requester can use any supported identifier supported by the IDP. Which is why SAML2Int recommends this setting. Because of that I'd suggest we default to `false` and no `NameIDPolicy` element (as recommended). Does that make sense?

PR #93, opened 2023-08-25
fix (saml): code refactoring

jimlambrt:
It may be easier to review this PR by commit.


saml/models/core/response_test.go
```diff
@@ -46,7 +46,6 @@ func Test_ParseResponse_ResponseContainer(t *testing.T) {
 
 	r.Equal(res.Destination, "http://localhost:8000/saml/acs")
 	r.Equal(res.ID, "saml-response-id")
-	r.Equal(res.InResponseTo, "saml-request-id")
```
jimlambrt, line 49: There doesn't appear to be `res.InResponseTo`

saml/models/core/response_test.go
```diff
@@ -71,14 +71,6 @@ var responseXMLStatus = `<?xml version="1.0" encoding="UTF-8"?>
    </saml2p:Status>
 </saml2p:Response>`
 
-func Test_ParseResponse_Status(t *testing.T) {
```
jimlambrt, line 74: there doesn't appear to be a `Status` field.

saml/response_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+// TODO: add the ability to sign requests, so we can write more complete unit tests
```
jimlambrt, line 14: This TODO is important, so we can write more complete unit tests.

saml/models/core/response_test.go
hcjulz: I wonder why. It was def. defined once. Will add it back, but thx for fixing this. I can add tests back once we have the status field added again.

hcjulz (approved):
👏 thx @jimlambrt LGTM

austingebauer (approved):
Thanks, Jim!

saml/response.go
```diff
@@ -110,55 +123,57 @@ func (sp *ServiceProvider) ParseResponse(
...
+				return nil, fmt.Errorf("%s: %w", op, ErrMissingAttributeStmt)
 			}
 		}
 	}
 
 	// TODO: transform gosaml2 response to core.Response
 	var result core.Response
-	xml.Unmarshal([]byte(samlResp), &result)
-
+	if xml.Unmarshal([]byte(samlResp), &result); err != nil {
```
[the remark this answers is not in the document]
austingebauer: Yes, thanks for the reminder. I'll put the PR up for this. Lots of changes on the Vault side recently and lost track 😅

PR #95, opened 2023-08-30
Add support for custom ACS URL in CreateAuthnRequest() and ParseResponse()

remilapeyre:
The URL can now be customized using `WithAssertionConsumerServiceURL()`
in both functions.

To validate the behavior I added a short test for `ServiceProvider.ParseResponse`.
It only checks the error to make sure `WithAssertionConsumerServiceURL()`
for now but can be extended in the future.

Also fix a docstring and gives the custom clock from `WithClock()` to the
internal parser.


saml/models/core/response_test.go
```diff
@@ -71,6 +71,14 @@ var responseXMLStatus = `<?xml version="1.0" encoding="UTF-8"?>
    </saml2p:Status>
 </saml2p:Response>`
 
+// func Test_ParseResponse_Status(t *testing.T) {
```
austingebauer, line 74: Any reason to keep commented out?

saml/models/metadata/sp_sso_descriptor_test.go
```diff
@@ -333,21 +333,21 @@ x5Ql0ejivIJAYcMGUyA+/YwJg2FGoA==
     </SPSSODescriptor>
 </EntityDescriptor>`
 
-func Test_SPSSODescriptor_KeyDescritpor(t *testing.T) {
-	r := require.New(t)
+// func Test_SPSSODescriptor_KeyDescritpor(t *testing.T) {
```
austingebauer, line 336: Same for this -- Any reason to keep commented out?

saml/response_test.go
```diff
@@ -83,3 +89,112 @@ func TestServiceProvider_ParseResponse(t *testing.T) {
...
+			} else {
+				require.ErrorContains(t, err, c.err)
+			}
+		})
+	}
+
+}
+
+// From https://www.samltool.com/generic_sso_res.php
+const responseSigned = `
```
austingebauer, line 163: It doesn't appear this is signed. `responseUnsigned`?

austingebauer (commented):
Couple of small suggestions and questions. Looks good otherwise!

saml/models/metadata/sp_sso_descriptor_test.go
austingebauer: Sorry, I see this doesn't build now. Makes sense to comment out.

saml/models/core/response_test.go
austingebauer: Also makes sense. Nevermind!

remilapeyre:
Thanks @austingebauer, all should be good now :)

PR #96, opened 2023-09-01
Adds Response type definition using gosaml2 type

austingebauer:
This PR creates a new SAML Response type definition using [`gosaml2.Types.Response`](https://github.com/russellhaering/gosaml2/blob/main/types/response.go#L36). Currently we're using `xml.Unmarshal` from this gosaml2 Response type to our own type which is less complete.

I'm open to other ideas here, including solutions that would require more involved work. If we don't end up going with a different strategy before this merges into main, we might want to add a disclaimer to the SAML module about the stability of the exported `Response` type.

The other changes in this PR are small cleanups.


austingebauer:
@remilapeyre still can't add you as reviewer so tagging you here.

saml/models/core/response.go
```diff
@@ -1,111 +1,9 @@
...
-	return a.Subject.NameID.Value
-}
-
-// Subject will return the subject format value.
-func (a *Assertion) GetSubjectFormat() string {
-	return string(a.Subject.NameID.Format)
-}
+// Response is a SAML Response element.
+// See 3.3.3 http://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf
+type Response types.Response
```
hcjulz, line 9: [for a separate PR] 
Nice! I think this will give us the ability to add methods to the response type. I think some functionality similar to the removed methods above to get the subject, issuer, and assertions, etc. They will come handy when implementing the ACS.

saml/models/core/response.go
```diff
@@ -1,111 +1,9 @@
 package core
 
 import (
-	"encoding/xml"
-	"time"
+	"github.com/russellhaering/gosaml2/types"
 )
 
-type Response struct {
```
hcjulz, line 8: I was thinking to leave the types here, but It makes sense to remove the types for now as they are incomplete and can lead to confusion.

hcjulz (approved):
🚀

saml/models/core/response.go
austingebauer: Indeed! I'll keep an eye out for any useful helpers we can introduce as I'll be using this for a different PR today :)

PR #104, opened 2023-09-12
saml: adds helpers for response assertions, subject, issuer, and attributes

austingebauer:
This PR adds a few helper methods for the SAML response. They allow for obtaining the assertions, attributes, and subject name ID. These helpers cover our usage of the response in the Vault SAML auth method and were useful to clean up additional nil checking that was previously required.

I will add tests before merging if this looks reasonable. I also want to restructure the tests in `response_test.go`.

### DIFF
--- a/saml/models/core/response.go
+++ b/saml/models/core/response.go
@@ -6,4 +6,53 @@ import (
 
 // Response is a SAML Response element.
 // See 3.3.3 http://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf
-type Response types.Response
+type Response struct {
+	types.Response
+}
+
+// Assertions returns the assertions in the Response.
+func (r *Response) Assertions() []Assertion {
+	assertions := make([]Assertion, 0, len(r.Response.Assertions))
+	for _, assertion := range r.Response.Assertions {
+		assertions = append(assertions, Assertion{Assertion: assertion})
+	}
+
+	return assertions
+}
+
+// Assertion is a SAML Assertion element.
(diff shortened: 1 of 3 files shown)


austingebauer:
@remilapeyre - I still can't tag you as a reviewer, so tagging you here 🙂

saml/models/core/response.go
```diff
@@ -6,4 +6,53 @@ import (
 
 // Response is a SAML Response element.
 // See 3.3.3 http://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf
-type Response types.Response
+type Response struct {
+	types.Response
```
austingebauer, line 10: I switched the approach here because I think it will be more flexible long term. For example, we can add additional fields to `Response` if needed or eventually migrate fields out of `types.Response`.

saml/models/core/response_test.go
```diff
@@ -81,12 +81,12 @@ func Test_ParseResponse_Assertion(t *testing.T) {
 	t.Parallel()
```
austingebauer, line 81: Going to restructure these tests and add more coverage if this PR looks reasonable.

jimlambrt (approved):
looks good.

hcjulz (approved):
Looks great! Thank you @austingebauer

saml/models/core/response_test.go
austingebauer: Done in https://github.com/hashicorp/cap/pull/104/commits/4655a4b4b988384b63f33b722ba7f3da55f9fbdd

PR #106, opened 2023-09-14
tests (saml): add tests for WithMetadataParameter(...) option

jimlambrt:


austingebauer (approved):
Thanks for adding these!

PR #107, opened 2023-09-14
downgrade ldap to v3.4.4

raymonstah:
Downgrading due to a race condition found in Vault's test.

We believe https://github.com/go-ldap/ldap/pull/388 introduced a race condition that's currently not present in v3.4.5


jimlambrt (approved):
ty!

PR #109, opened 2023-09-18
update go-ldap to v3.4.6

raymonstah:
go-ldap v3.4.6 has been [released](https://github.com/go-ldap/ldap/releases/tag/v3.4.6) which includes the fix for the race condition introduced in v3.4.5.


jimlambrt (approved):
ty!

PR #128, opened 2024-02-01
Add support for multiple JWKS urls

johnlanda:
This PR is intended as a building block to add support for multiple JWKS URLs in a single auth method. 

Main changes:
* Update the `Validator` to take multiple key sets.
* Update the validation logic to loop over keysets and attempt to verify JWT with any available

### DIFF
--- a/jwt/jwt.go
+++ b/jwt/jwt.go
@@ -8,7 +8,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"strings"
 	"time"
 
 	"github.com/go-jose/go-jose/v3"
@@ -19,27 +18,26 @@ import (
 // for validating the "nbf" (Not Before) and "exp" (Expiration Time) claims.
 const DefaultLeewaySeconds = 150
 
-type Validator interface {
-	Validate(ctx context.Context, token string, expected Expected) (map[string]interface{}, error)
-	ValidateAllowMissingIatNbfExp(ctx context.Context, token string, expected Expected) (map[string]interface{}, error)
-}
-
 // Validator validates JSON Web Tokens (JWT) by providing signature
 // verification and claims set validation.
-type validator struct {
-	keySet KeySet
+type Validator struct {
+	keySets []KeySet
 }
 
-var _ Validator = (*validator)(nil)
-
 // NewValidator returns a Validator that uses the given KeySet to verify JWT signatures.
-func NewValidator(keySet KeySet) (Validator, error) {
-	if keySet == nil {
-		return nil, errors.New("keySet must not be nil")
+func NewValidator(keySets ...KeySet) (*Validator, error) {
+	if len(keySets) <= 0 {
+		return nil, errors.New("must provide at least one key set")
 	}
 
-	return &validator{
-		keySet: keySet,
+	for _, keySet := range keySets {
+		if keySet == nil {
+			return nil, errors.New("keySet must not be nil")
+		}
+	}
+
+	return &Validator{
+		keySets: keySets,
 	}, nil
 }
 
@@ -103,7 +101,7 @@ type Expected struct {
 //     and "exp" (Expiration Time) claims and after the time given by the "iat"
 //     (Issued At) claim, with configurable leeway. See Expected.Now() for details
 //     on how the current time is provided for validation.
-func (v *validator) Validate(ctx context.Context, token string, expected Expected) (map[string]interface{}, error) {
+func (v *Validator) Validate(ctx context.Context, token string, expected Expected) (map[string]interface{}, error) {
 	return v.validateAll(ctx, token, expected, false)
 }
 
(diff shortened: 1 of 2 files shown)


jwt/jwt.go
```diff
@@ -227,6 +235,73 @@ func (v *Validator) validateAll(ctx context.Context, token string, expected Expe
...
+	}
+
+	return &multiValidator{
+		validators: validators,
+	}, nil
+}
+
+func (m *multiValidator) Validate(ctx context.Context, token string, expected Expected) (map[string]interface{}, error) {
+	for _, v := range m.validators {
+		claims, err := v.Validate(ctx, token, expected)
```
johnlanda, line 271: I think that we can optimize this if for each `keySet` we expose some way to get a `[]string` which contains all `kid` for the keys in the set. 

Then rather than attempting each validator, we can define candidates as the validators which contain the `kid` for the incoming JWT in their `keySet`. I think we would still probably want to check all sets rather than short circuit on the first hit in case of a collision. 

Let me know what you think @austingebauer

jwt/jwt.go
```diff
@@ -227,6 +235,73 @@ func (v *Validator) validateAll(ctx context.Context, token string, expected Expe
...
+		}
+
+		return claims, nil
+	}
+
+	return nil, errors.New("no key set was able to verify the token signature")
+}
+
+func (m *multiValidator) ValidateAllowMissingIatNbfExp(ctx context.Context, token string, expected Expected) (map[string]interface{}, error) {
+	for _, v := range m.validators {
```
johnlanda, line 288: I will pull this logic into a helper function on the next commit as well since the only difference is the `v.Validate` vs `v.ValidateAllowMissingIatNbfExp`. I'll make it so that the helper takes a delegate function and then this should all get a bit smaller.

jwt/jwt.go
```diff
@@ -18,19 +19,26 @@ import (
...
-type Validator struct {
+type validator struct {
 	keySet KeySet
 }
 
+var _ Validator = (*validator)(nil)
+
 // NewValidator returns a Validator that uses the given KeySet to verify JWT signatures.
-func NewValidator(keySet KeySet) (*Validator, error) {
+func NewValidator(keySet KeySet) (Validator, error) {
```
austingebauer, line 36: I think the introduction of `Validator` isn't necessarily wrong, but I know we've tried to follow not returning interfaces with some exceptions that we think follow [generality](https://go.dev/doc/effective_go#generality). I could see the argument that this is another case for that.

I was thinking it'd be simpler and less code / api surface to make `NewValidator` a variadic function.

Something like: 
```go
func NewValidator(keySet ...KeySet) (*Validator, error) { ... }
```

I _think_ it would be roughly ~10 lines of code (tests excluded) with that approach to support multiple key sets. Happy to hear reasons why the approach here is better. Lmk!

johnlanda: That is a great idea. My original intention was to minimize touching any validator logic by adding the abstraction layer with the interface. I'll have a go at the refactoring for the variadic function instead. 

I think variadic makes sense in this case since I do not expect that there will be another `Validator` implementation in the future - in which case I would lean towards the interface for extensibility.

jwt/jwt_test.go
```diff
@@ -574,27 +584,41 @@ func TestNewValidator(t *testing.T) {
...
 			name: "new validator with nil keySet",
 			args: args{
-				keySet: func() KeySet {
+				keySets: func() []KeySet {
 					return nil
 				},
 			},
 			wantErr: true,
 		},
+		{
```
austingebauer, line 604: Could add one more case for the nil key set in the list of key sets. Thanks for adding that validation!

```
{
	name: "new validator with nil keySet in keySets",
	args: args{
		keySets: func() []KeySet {
			ks, err := NewJSONWebKeySet(context.Background(),
				"https://issuer.com/"+wellKnownJWKS, "")
			require.NoError(t, err)
			return []KeySet{ks, nil}
		},
	},
	wantErr: true,
},
```

austingebauer (approved):
Thanks, @johnlanda! LGTM after the changes.

jwt/jwt.go
austingebauer: That seems like a nice optimization! Can you elaborate on the collision case you were thinking of? I'd be open to trying something like this in a future PR 👍

jwt/jwt.go
```diff
@@ -21,17 +21,23 @@ const DefaultLeewaySeconds = 150
 // Validator validates JSON Web Tokens (JWT) by providing signature
 // verification and claims set validation.
 type Validator struct {
-	keySet KeySet
+	keySets []KeySet
 }
 
 // NewValidator returns a Validator that uses the given KeySet to verify JWT signatures.
```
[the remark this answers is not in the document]
johnlanda: Ack. Added in 6ee096b4b98151bc6f228d98a4af330a0df75034

jwt/jwt.go
johnlanda: For the collision case: since the JWKS at each `jwks_url` are independent it would be possible that there are two keys in different JWKS that have the same key id `kid`. 

If we were to do something like the following 

```
// An incoming JWT with a `kid` claim for the key id that corresponds to it
var neededKID := ourJWT.kidClaim // making up the syntax here

// jwks_url -> set of key ids in the set
var knownKeys map[string][]string

// assume some other routine keeps the known keys up to date - can probably tie into caching
for jwks_url, keys := range knownKeys {
  if slices.Contains(keys, neededKID) {
    // Try the actual verification using this keyset since we think it could have the right key
    ...
  }
}
```

If we were to break from the loop over the `knownKeys` early there would be a small chance that there is another keyset which contains a key with the same identifier and we could falsely claim that we failed to verify the JWT.

I think that this could save us on CPU over the current approach since we don't need to attempt to do the verification with each keyset. Only the ones that are known to have a key with a matching key id.

PR #132, opened 2024-03-27
Add case insensitive user attribute keys configs for LDAP

jasonodonnell:
Vault relies on [case insensitive user attribute keys](https://github.com/hashicorp/vault-enterprise/blob/main/sdk/helper/ldaputil/config.go#L302-L304) for mapping user attributes to entity alias metadata. This adds configs to allow the callers to specify how the attribute keys should be returned.


fairclothjm (approved):

ldap/options.go
```diff
@@ -75,6 +76,18 @@ func WithUserAttributes() Option {
 	}
 }
 
+// WithLowerUserAttributeKeys returns a User Attribute map where the keys
+// are all cast to lower case. This is neccessary for some clients, such as Vault,
```
jimlambrt, line 80:
```suggestion
// are all cast to lower case. This is necessary for some clients, such as Vault,
```

ldap/client.go
```diff
@@ -273,7 +273,11 @@ func (c *Client) Authenticate(ctx context.Context, username, password string, op
 			return nil, fmt.Errorf("%s: failed to get user attributes: %w", op, err)
 		}
 		for _, a := range attrs {
-			userAttrs[a.Name] = a.Vals
+			name := a.Name
```
jimlambrt, line 276: Mind adding WithLowerUserAttributeKeys to the list of supported optins in the
func's godoc?

jimlambrt (changes requested):
looks good overall.  couple minor things.

ldap/client.go
jasonodonnell: Good call, done

jimlambrt (approved):
ty!

PR #134, opened 2024-03-28
Add empty anonymous group search configs

jasonodonnell:
Anonymous group searches can be rejected by some LDAP servers if they contain a userDN. I've added a toggle here that let's the caller to specify unauthenticated binds for anonymous group searches should exclude a DN.


jimlambrt (approved):
ty!

PR #135, opened 2024-04-03
Add ability to the SAML test provider to create signed SAML responses

hcjulz:
This PR adds the ability to the SAML test provider to create signed SAML responses. It uses the [goxmldsig](https://github.com/russellhaering/goxmldsig) library for this purpose, which is also used by the underlaying `gosaml2` library used in `cap/saml`.  

Before, this library used `samltest.id` (a test SAML Identity provider), which doesn't exist anymore. The tests for parsing SAML responses have been updated to use the test provider.

I'll open a follow up PR to improve the test SAML identity provider in `cap/saml`. Areas to improve:

- Provide more options to modify the SAML response, such that we can get rid of any hard coded SAML response XML strings.
- Create the IDP Metadata programatically with options.

### DIFF
--- a/saml/response_test.go
+++ b/saml/response_test.go
@@ -9,13 +9,14 @@ import (
 	"testing"
 	"time"
 
-	"github.com/hashicorp/cap/saml"
-	"github.com/hashicorp/cap/saml/models/core"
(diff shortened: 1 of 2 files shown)


jimlambrt (approved):
ty!

PR #139, opened 2024-08-13
Add Endpoint() api call to oidc.Provider

DrDaveD:
This adds an `Endpoint()` api call to `oidc.Provider` which simply invokes the underlying coreos/go-oidc `Endpoint()` api.  That returns an `oauth2.Endpoint` object which since coreos/go-oidc#365 includes the `DeviceAuthURL` in addition to the `AuthURL` and `TokenURL` values read by oidc discovery.  This enables simplifying applications that need to use those URLs and avoids them having to separately discover that information, for example hashicorp/vault-plugin-auth-jwt#131 which adds the device flow to jwt/oidc authentication.

### DIFF
--- a/oidc/provider.go
+++ b/oidc/provider.go
@@ -826,3 +826,9 @@ func unmarshalRespJSON(r *http.Response, body []byte, v interface{}) error {
 	}
 	return fmt.Errorf("%s: expected Content-Type = application/json, got %q and could not unmarshal it as JSON: %w", op, ct, err)
 }
+
+// Endpoint returns the oauth2 Endpoint information that the go-oidc
+// provider already knows about from its probe of the Discovery URL
+func (p *Provider) Endpoint() oauth2.Endpoint {
+        return p.provider.Endpoint()
+}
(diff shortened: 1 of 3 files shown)


jimlambrt (changes requested):
Ty for the PR. Next, perhaps consider opening an issue first where we can discuss the proposed feature/change and get alignment on how to proceed? 

With that said, adding this seems reasonable as long as the contribution covers all the necessary changes and not just a one-off additional provider func to dig out the oauth2.Endpoint if the provider returned it in its discovery reponse.  To that end I have a few comments/questions/suggestions:

* How would this work if the provider doesn't support discovery and is configured via the `WithProviderConfig(...)` option?  
* I think, we need a solution that supports both discovery and the `WithProviderConfig` option.  If we do that we can likely return a copy of the provider's `oidc.ProviderConfig` here which makes it consistent with the current implementation which wraps the underlying oauth2.Endpoint with it's own type.
* We need unit tests for all changes, so we need to add support for the device AuthURL to the oidc.TestProvider

DrDaveD:
Thanks for the quick response.  I had started to add a DeviceAuthURL field to ProviderConfig and to DiscoveryInfo but it appeared to be opening a big can of worms so I thought I would try first just this most simple of changes, especially since that was exactly what I needed and I don't have any idea if anyone else would use the other changes.  Next I will try to do what you ask.

DrDaveD:
Replaced by #140

PR #140, opened 2024-08-14
Add support for DeviceAuthURL and a Provider.Config() API

DrDaveD:
The go-oidc library added support for `DeviceAuthUR`L in coreos/go-oidc#365.  This adds it also to the cap library and also makes a `ProviderConfig` structure available through a new API call `Provider.Config()` so an application like the one added in hashicorp/vault-plugin-auth-jwt#131 that needs those URLs does not need to discover them itself.  Note that it does not include JWKSURL because that is not available through the go-oidc API.

At the same time, this adds `DeviceAuthURL` to the `DiscoveryInfo` API.

- Replaces #139


jimlambrt:
@DrDaveD I'll try to get you some review feedback by end of this week; baring unforeseen circumstances.

DrDaveD:
@jimlambrt I suppose those unforeseen circumstances occurred?  Can you get that review feedback soon?

PR #145, opened 2024-11-29
Enhancing signature validation in SAML Response #144

himran92:
The SAML protocol allows signing of a Response, its Assertion(s), neither, or both. Since Assertion(s) are sub-elements of a Response, they are signed if the Response is signed. Today we are depending on the gosaml2 for signature validation. It only checks only one or the other be signed.
For enhancing security, we would like to give user the options to sign just assign response or assertion or both. 

Changes in PR include:
- adding two new options for parse response 
- enhancing SAML response to validate either response or assertion or both according to the given options
- extending provider to allow signing response or assertion as controlled by parameters
- extending tests to cover these cases

Points before/after merging: 
- We will only merge once the affiliated PR for vault SAML auth plugin is also approved. 
- Once approved, we will work with repo owners on cutting a new release. Since the last release was cut, i see this current PR for SAML and this other [PR](https://github.com/hashicorp/cap/commit/67f22c4214b283f3c1108ee682b7770343b2d14a#diff-f83b7ecdf202dd11720aa6089cdbf52761b3eaf5723a8cce3a2c3d201f8648a4R96) for oidc as the impacting change. As the other PR is for _oidc_, it seems it will only impact vault-plugin-auth-jwt when its updated to use the latest cap release. Checked with oidc PR author that it is a backward compatible change. 
EDIT: there is another [PR](https://github.com/hashicorp/cap/pull/146) merged in cap related to `Adds an option to enable sAMAccountname logins when upndomain is set`


saml/response.go
```diff
@@ -12,6 +12,7 @@ import (
 
 	"github.com/jonboulle/clockwork"
 	saml2 "github.com/russellhaering/gosaml2"
+	"github.com/russellhaering/gosaml2/types"
```
hcjulz, line 15: Can we use the `cap/saml` types instead? https://github.com/hashicorp/cap/blob/main/saml/models/core/response.go#L10-L14

saml/response.go
```diff
@@ -245,3 +254,21 @@ func parsePEMCertificate(cert []byte) (*x509.Certificate, error) {
 
 	return x509.ParseCertificate(block.Bytes)
 }
+
+func validateSignature(response *types.Response, op string) error {
+	// validate child object assertions
+	for _, assert := range response.Assertions {
```
hcjulz, line 260: gosaml2 requires either the response or the assertions to be signed. It doesn't require both. 

My qustion would be if we need to sign both because signature inheritance applies (if the parent object is signed, the child object is included)?
Having both objects signed makes sense in some instances where the service provider could retrieve the assertion outside of a SAML Response, but that's not the case here.

saml/response.go
```diff
@@ -151,6 +152,14 @@ func (sp *ServiceProvider) ParseResponse(
 		}
 	}
 
+	if !opts.skipSignatureValidation {
```
hcjulz, line 155: Can we make this fully optional? Something like `opts.ResponseSigned` or `opts.ResponseAndAssertionsSigned`?

hcjulz (commented):
Hey @himran92 👋 Thx a lot for opening this PR.

A general comment:
I think `gosaml2` follows the SAML spec and it is not a requirement to sign both Response and Assertions. If the Reponse is signed signature inheritance applies, and if the assertions are signed it leaves only the Response unsigned. The fields in the Response object are not really an issue if tempered. The reason is mainly because relevant fields like the `InResponseTo` field is  (should be) checked on the service provider. The cap/saml implementation by default requires the `InResponseTo` field to be checked and explicitly prefixes the option  to leave it out with `Insecure`.  If additional protection is required.

If we, however, still want to have the entire response to be signed, I'd suggest to make this optional. If this is forced it could lead to be incompatible with identity providers that do not provide the option to signed the full response.

I left a few comments/questions.

saml/response.go
himran92: Yes agree that we need to make this optional. 
I had a discussion internally and moved the PR to draft again to bring the optional field change + test the e2e flow with vault.

saml/response.go
hcjulz: I did some further digging on this. Most SAML IDPs (e.g. Entra ID and Okta) do provide the option to sign both: `Response` and `Assertions`. However, it looks like there are indeed some IDPs that do not provide the option, such as `Auth0`. This is from their application settings:

```
signResponse (bool): Whether or not the SAML Response should be signed. By default the SAML Assertion will be signed, but not the SAML Response. If true, SAML Response will be signed instead of SAML Assertion.
```

On your edit above in the description:
> After few major version, the vault variable will be removed and user will always have to sign both.

Making both a requirement might lead to incompatibilities with some IDPs (like Auth0 for example). So I wonder if supporting all 3 options would be the ideal case:
- Either `Response` or `Assertion` has to be signed
- `Response` has to be signed
- Both have to be signed.

saml/response.go
```diff
@@ -19,19 +19,21 @@ import (
...
-	skipRequestIDValidation          bool
-	skipAssertionConditionValidation bool
-	skipSignatureValidation          bool
-	assertionConsumerServiceURL      string
+	clock                                   clockwork.Clock
+	skipRequestIDValidation                 bool
+	skipAssertionConditionValidation        bool
+	skipSignatureValidation                 bool
+	assertionConsumerServiceURL             string
+	requireSignatureForResponseAndAssertion bool
```
himran92, line 27: For `requireSignatureForResponseAndAssertion` :
On the vault usage side i opted for a name `EnableStrictResponseSignatureValidation` as that sounded more user friendly for a user exposed setting. The description will share the impact of the setting. Whereas kept it `requireSignatureForResponseAndAssertion` on cap side thinking of it as more of an internal lib so can have any name. 

Any preference of using `EnableStrictResponseSignatureValidation`  here?

hcjulz: I don't have a strong preference. Based on my [comment](https://github.com/hashicorp/cap/pull/145#discussion_r1867487684) about having all 3 options signature validation, it would just be important to think about the general naming pattern:

1. Validate Response And Assertion Signatures = `EnableStrictSignatureValidation`
2. Validate Response Signature = `EnableResponseSignatureValidation`?

Alternatively, we could use a more explicit naming pattern:

1. `ValidateResponseSignature`
3. `ValidateResponseAndAssertionSignatures`

I'm ok either way.

saml/response.go
himran92: done

saml/response.go
himran92: Changes done to add separate options for all signature validation. Really appreciate for you finding the info on Auth0 as IDP.

saml/response.go
himran92: Made all settings optional.

saml/response.go
himran92: Updated the name of the setting.

saml/response.go
himran92: Also btw the options i kept are: 
- assertion is signed
- response is signed
- both are signed
Did not include the `either` option as i thought user should makes a known decision relative to their IDP. Let me know if you think otherwise.

hcjulz: Hey @himran92 sorry for the late response and thx a lot for addressing my comments 🙇 

Sounds great. The options makes totally sense. 

The default would still be `either`, correct?

himran92: I am catching up after vacations :). 

yes correct. 
The default where neither of these options is provided is still `either`. 
This is given by these [unit tests](https://github.com/hashicorp/cap/blob/d857fbab8bb1108868cdd6b03433293f72304809/saml/response_test.go#L68-L81).

saml/response.go
```diff
@@ -87,15 +121,21 @@ func (sp *ServiceProvider) ParseResponse(
...
 	switch {
 	case sp == nil:
 		return nil, fmt.Errorf("%s: missing service provider %w", op, ErrInternal)
 	case samlResp == "":
 		return nil, fmt.Errorf("%s: missing saml response: %w", op, ErrInvalidParameter)
 	case requestID == "":
 		return nil, fmt.Errorf("%s: missing request ID: %w", op, ErrInvalidParameter)
+	case opts.skipSignatureValidation && callValidateSignature:
+		return nil, fmt.Errorf("%s: option `skip signature validation` cannot be true with any validate signature option : %w", op, ErrInvalidParameter)
+	case multipleSignatureOptionEnabled(opts.validateResponseAndAssertionSignatures, opts.validateResponseSignature, opts.validateAssertionSignature):
```
kpcraig, line 136: I think I see discussion on these options already in the discussion, so I'll defer to `cap` people on the ergonomics/patterns, but if we need to check if the options are sensible, would it be better to reshape the options so that any pattern is valid? (for example, can we remove the "both" boolean and have `validateResponse` and `validateAssertion` both be true instead?

If there is a need/desire to have an explicit "both" option mutually exclusive with the other two, would it be worth turning it into a single field set to `const` values, e.g. 
```
const (
   NoSignatureValidation = iota
   ValidateAssertionSignature
   ValidateResponseSignature
   ValidateResponseAndAssertionSignature
)
```

I understand we wouldn't want to remove the `InsecureSkipValidation` functions, for compatibility reasons.

himran92: @hcjulz  Do you have any recommendation for above w.r.t cap?

I like Kay point to remove the `both` option and ask user to set individual option instead.

himran92: @mgaffney  just randomly tagging you as you are mentioned as the repo owner. 
Would you be able to assist in the review as Julian is out till end of the year, thanks

Let me know if you prefer/recommend someone else.

himran92: Mike was not familiar with this piece of code..
tagging @jimlambrt  for review for Pr.

jimlambrt: I think, removing `ValidateResponseAndAssertionSignatures` makes for a better
dev experience.

jimlambrt (changes requested):
Ty for the tests!

saml/response.go
himran92: PR is updated now with only keeping two options

jimlambrt (approved):
Ty!

PR #146, opened 2024-12-06
Adds an option to enable sAMAccountname logins when upndomain is set

kwagga:
Active Directory allows LDAP binds as `userprincipalname@upndomain` as well as `samaccountname@updomain`. 

With the current LDAP filter `samaccountname` logins fail when the `upndomain` configuration parameter is set, since the filter only only checks for `userprincipalname=username@updomain`.

This PR provides an `enable_samaccountname_login` option that can be set in the LDAP Authentication method. This will cause the LDAP user search filter to match either `userprincipalname` or `samaccountname` attributes instead of just the `userprincipalname` when the `upndomain` configuration parameter is set.


jimlambrt (changes requested):
Perhaps some unit tests?

kwagga:
A unit test is a great idea! I've attempted adding a test based on [success-with-anon-bind-upn-domain](https://github.com/hashicorp/cap/blob/main/ldap/client_exported_test.go#L326-L339) by adding the  `EnableSamaccountnameLogin: true,` client option. The test is successful, but since the `eve` user does not have a `sAMAccountname` attribute and value the ldap filter isn't tested. Would such a test be sufficient?

kwagga:
I've added a few unit tests that sets `EnableSamaccountnameLogin: true`.

Ideally I would have wanted to test the ldap filter by passing `sAMAccountname=evesam@example.com`. That would require changes to [gldap.testdirectory](https://github.com/jimlambrt/gldap/tree/main/testdirectory). Adding a `sAMAccountname` attribute is possible but the [gldap.testdirectory](https://github.com/jimlambrt/gldap/tree/main/testdirectory) builds a **D**istinguished **N**ame `userPrincipalName=eve@example.com,ou=people,dc=example,dc=org` object which is rather strange. I expected a DN of `cn=eve,ou=people,dc=example,dc=org` with a `UserPrincipalName` object instead.

jimlambrt:
I did some digging and I agree that gldap needs some changes.  I'm willing to accept this as is for now and then write a proper unit test in a future PR.

jimlambrt (approved):
See comment: we need better unit tests in a future PR

PR #155, opened 2025-02-25
OIDC private key JWT / client assertion

gulducat:
Hello! I'm a Nomad engineer, implementing "Private Key JWT" a.k.a. `client_assertion` to our OIDC authentication options, as an alternative to Client Secret auth (although technically, one may sign an asertion JWT with a client secret instead of a private key. this supports that, too).

Reference: https://oauth.net/private-key-jwt/

This comprises two main components:
 1. Creating a JWT with the right shape for providers to verify
    * I've added a new `clientassertion` package to isolate this logic
    * This is the bulk of the code in this PR, with tests
 2. Adding the JWT to the auth request for the client to send
    * I opted to accept the JWT as a string, which can be generated using the new package
    * This is mostly simple plumbing, and some tests

\# 1 does not necessarily need to be in this library, but it seemed reasonable to me. # 2 is required for Nomad's existing usage of `hashicorp/cap` (example in Nomad repo [here](https://github.com/hashicorp/nomad/blob/v1.9.6/nomad/acl_endpoint.go#L2737))

Nomad will use this sign client assertion JWTs with any of:
* Nomad's internal keyring (RS256), which includes a JWKS endpoint for providers that support that
* A user-provided RSA key (RS256)
* A client secret HMAC (HS256)

---

Here is a stripped-down example usage of it:

```go
assertion, err := clientassertion.New(clientID, audience,
	clientassertion.WithRSAKey(rsaKey, "RS256"),
	clientassertion.WithKeyID(keyID),
)
die(err)
token, err := assertion.SignedToken()
die(err)
req, err := capOIDC.NewRequest(expiration, redirect,
	capOIDC.WithClientAssertionJWT(token),
)
// and so on as usual
```

---

Test coverage of the `clientassertion` package is high, near 100%. I added tests to the outer `oidc` library as best I could figure out how it's arranged.

Hopefully this may also make it easier for folks to implement this feature in other products, as they see fit.


oidc/clientassertion/client_assertion.go
```diff
@@ -0,0 +1,206 @@
...
+	clientID string
+	audience []string
+	headers  map[string]string
+
+	// for signer
+	alg jose.SignatureAlgorithm
+	// key may be any key type that jose.SigningKey accepts for its Key
+	key any
+	// secret may be used instead of key
+	secret string
```
johanbrandhorst, line 63: Any constraints on the string length or format? Sounds like go-jose [requires this to be a HMAC key](https://pkg.go.dev/github.com/go-jose/go-jose/v4#SigningKey), do we want to use the input key to derive a HMAC key (perhaps with HKDF)?  We could potentially do it in the `WithSecret` function directly so that all assignments to this value are guaranteed to be of the right format.

oidc/clientassertion/client_assertion.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+	if c.alg == "" {
+		errs = append(errs, ErrMissingAlgorithm)
+	}
+	if c.key == nil && c.secret == "" {
+		errs = append(errs, ErrMissingKeyOrSecret)
+	}
+	if c.key != nil && c.secret != "" {
+		errs = append(errs, ErrBothKeyAndSecret)
+	}
```
johanbrandhorst, line 98: We should validate that the size of the secret matches the algorithm if `c.secret` is used.

oidc/clientassertion/client_assertion.go
```diff
@@ -0,0 +1,206 @@
...
+		Audience:  c.audience,
+		Expiry:    jwt.NewNumericDate(now.Add(5 * time.Minute)),
+		NotBefore: jwt.NewNumericDate(now.Add(-1 * time.Second)),
+		IssuedAt:  jwt.NewNumericDate(now),
+		ID:        id,
+	}
+}
+
+// Options configure the ClientAssertion
+type Option func(*ClientAssertion)
```
johanbrandhorst, line 173: In Boundary nowadays we usually prefer the format
```suggestion
type Option func(context.Context, *ClientAssertion) error
```
For option types, so that errors can happen in the option, and context for cancellation/values. The context isn't necessary, but the error is nice, that allows validation to happen in the option, which can then propagate up through the function that consumes the option.

oidc/clientassertion/client_assertion.go
```diff
@@ -0,0 +1,206 @@
...
+
+// WithClientSecret sets a secret and algorithm to sign the JWT with
+func WithClientSecret(secret string, alg string) Option {
```
johanbrandhorst, line 176: Can't we make this a `jose.SignatureAlgorithm`?
```suggestion
func WithClientSecret(secret string, alg jose.SignatureAlgorithm) Option {
```
Also, given that this is taking a string, I assume we only allow `HSXXX` signing algorithms? Where do we validate that?

oidc/clientassertion/client_assertion_test.go
```diff
@@ -0,0 +1,287 @@
...
+				WithHeaders(map[string]string{"xtra": "headies"}),
+			},
+		},
+		{
+			name:     "invalid alg",
+			claimKey: pub,
+			opts: []Option{
+				WithRSAKey(key, "ruh-roh"),
+			},
+			err: jose.ErrUnsupportedAlgorithm,
```
johanbrandhorst, line 222: Do we want to make this error part of our public API? Maybe it should be intercepted and converted to one from this package?

oidc/clientassertion/doc.go
```diff
@@ -0,0 +1,13 @@
...
+// in OIDC client_assertion requests, A.K.A. private_key_jwt. reference:
+// https://oauth.net/private-key-jwt/
+//
+// Example usage:
+//
+// cass, err := clientassertion.New("client-id", []string{"audience"},
+// 	clientassertion.WithRSAKey(rsaPrivateKey, "RS256"),
+// 	clientassertion.WithKeyID("jwks-key-id-or-x5t-etc"),
+// )
+// jwtString, err := cass.SignedToken()
```
johanbrandhorst, line 13: What do you think about adding an `example_test.go` file with this code (and error handling)?

oidc/provider.go
```diff
@@ -306,6 +307,12 @@ func (p *Provider) Exchange(ctx context.Context, oidcRequest Request, authorizat
 	if oidcRequest.PKCEVerifier() != nil {
 		authCodeOpts = append(authCodeOpts, oauth2.SetAuthURLParam("code_verifier", oidcRequest.PKCEVerifier().Verifier()))
 	}
+	if oidcRequest.ClientAssertionJWT() != "" {
+		authCodeOpts = append(authCodeOpts,
+			oauth2.SetAuthURLParam("client_assertion_type", cass.ClientAssertionJWTType),
+			oauth2.SetAuthURLParam("client_assertion", oidcRequest.ClientAssertionJWT()),
+		)
+	}
```
johanbrandhorst, line 315: Is this exclusive with any other option? I'm not too familiar with this OAuth2 pattern.

oidc/request.go
```diff
@@ -83,6 +83,12 @@ type Request interface {
 	// See: https://tools.ietf.org/html/rfc7636
 	PKCEVerifier() CodeVerifier
 
+	// ClientAssertionJWT optionally specifies a signed JWT to be used in a
+	// client_assertion token request.
+	//
+	// See: https://oauth.net/private-key-jwt/
+	ClientAssertionJWT() string
```
johanbrandhorst, line 90: Could we make this return something more structured than a string? Since this is an interface, I think it might be nice to make this stronger typed. Could we use the *`clientassertion.ClientAssertion` type?

oidc/clientassertion/client_assertion.go
```diff
@@ -0,0 +1,206 @@
...
+	ErrMissingAlgorithm   = errors.New("missing signing algorithm")
+	ErrMissingKeyOrSecret = errors.New("missing private key or client secret")
+	ErrBothKeyAndSecret   = errors.New("both private key and client secret provided")
+	// if these happen, either the user directly instantiated &ClientAssertion{}
+	// or there's a bug somewhere.
+	ErrMissingFuncIDGenerator = errors.New("missing IDgen func; please use New()")
+	ErrMissingFuncNow         = errors.New("missing now func; please use New()")
+)
+
+// New sets up a new ClientAssertion to sign private key JWTs
```
tgross, line 33: It doesn't look like we actually have any callers for `clientassertion.New` here in the `cap` library. And `New` is the only non-test code path that calls `Validate`. I do get that we pulled this out to a package for testability.

But from the perspective of an application consuming this API, I'd question why my application needs to call `clientassertion.New` instead of passing the appropriate configuration values to `oidc.NewConfig`. It seems like we're introducing more surface area here (and therefore more ways to get it wrong in the application).

oidc/clientassertion/client_assertion.go
```diff
@@ -0,0 +1,206 @@
...
+		opt(a)
+	}
+	if err := a.Validate(); err != nil {
+		return nil, fmt.Errorf("new client assertion validation error: %w", err)
+	}
+	return a, nil
+}
+
+// ClientAssertion signs a JWT with either a private key or a secret
+type ClientAssertion struct {
```
johanbrandhorst, line 52: This name stutters a bit (`clientassertion.ClientAssertion`). What do you think about calling this `JWT` instead? Since the only exposed functions are `Validate` and `SignedToken`, we could just rename `SignedToken` to `Serialize` or something.

oidc/clientassertion/client_assertion.go
```diff
@@ -0,0 +1,206 @@
...
+		sKey.Key = []byte(c.secret)
+	}
+	if c.key != nil {
+		sKey.Key = c.key
+	}
+
+	sOpts := &jose.SignerOptions{
+		ExtraHeaders: make(map[jose.HeaderKey]interface{}, len(c.headers)),
+	}
+	// note: extra headers can override "kid"
```
johanbrandhorst, line 147: Do we want to prevent this?

johanbrandhorst (changes requested):
Have you been able to integration test this against any third party OAuth2 servers to see that it works as expected?

oidc/clientassertion/client_assertion.go
gulducat: My read of `oidc.Config` is to configure the `Provider`, which seems to include some minimum required features of OIDC. Other things (like PKCE, which I followed for my implementation) are more features of individual parts of the process. For client assertions (as with PKCE), it's part of a `Request`.

I could put more of this logic into the `oidc.Option`, which is passed in `oidc.NewRequest` -- currently `oidc.WithClientAssertionJWT()` just takes a `string`, but it could instead take all the things that this thing takes, and produce the string itself internally. I'm just not sure how much that buys us, and I figured leaving this open would allow applications to produce the JWT however they may want, making this package convenient but optional.

@johanbrandhorst suggests below that we could pass the `ClientAssertion` (perhaps renamed to `JWT` to avoid stuttery package.struct names) instead of a string, then the option or Request constructor could do the signing (if I'm understanding that suggestion right). That would make this extra package a lot less optional for this feature, but definitely harder to hold wrong.

Maybe an in-between would be to accept an interface, which `ClientAssertion`/`JWT` implements, so folks could build their own, if they wanted. Harder to hold wrong, but still flexible.

oidc/clientassertion/client_assertion.go
gulducat: That does seem a nice bit of extra validation, yep, to maybe get that error before it gets sent through jose.

oidc/clientassertion/client_assertion.go
gulducat: I don't think it actually hurts anything to let folks do this, but it could potentially cause confusion?

oidc/clientassertion/client_assertion.go
gulducat: I dig this. I was following the `oidc` package convention, but an `error` here could save some logic in `Validate`.

oidc/clientassertion/client_assertion.go
gulducat: I've been thinking "let jose complain about it" but yeah, we could bail out earlier in the process if we checked ourselves.

oidc/clientassertion/doc.go
gulducat: Can't believe I forgot about that! Will do.

oidc/provider.go
gulducat: I know that it at least plays nicely with PKCE. If the assertion is signed by the ClientSecret, then it sorta re-contextualizes that param, but otherwise I don't think it's exclusive of other things? Perhaps I'm ignorant of some nuance here though (there is a lot of nuance around OIDC, I've learned).

gulducat: I found this particular implementation pattern, using `oauth2.SetAuthURLParam`, in this issue where someone requested it in the `oauth2` repo here: https://github.com/golang/oauth2/issues/744#issuecomment-2411878895

oidc/request.go
gulducat: Yeah, I'm going to take a swing at this, because it would make it harder to hold incorrectly.

gulducat:
Regarding your top-level comment @johanbrandhorst 

> Have you been able to integration test this against any third party OAuth2 servers to see that it works as expected?

Yep! I have tested it using KeyCloak (easy to run on laptop) and Azure EntraID (what our customer who requested this has said they use). I wanted to test with Auth0 too, but this feature is an enterprise feature over there, and I didn't want to jump through those hoops. I haven't checked Okta or others.

Happy to take requests!

oidc/clientassertion/client_assertion.go
tgross: > @johanbrandhorst suggests below that we could pass the ClientAssertion (perhaps renamed to JWT to avoid stuttery package.struct names) instead of a string, then the option or Request constructor could do the signing (if I'm understanding that suggestion right). That would make this extra package a lot less optional for this feature, but definitely harder to hold wrong.
>
> Maybe an in-between would be to accept an interface, which ClientAssertion/JWT implements, so folks could build their own, if they wanted. Harder to hold wrong, but still flexible.

Yeah, either of those options sounds pretty solid.

oidc/clientassertion/client_assertion_test.go
johanbrandhorst: Haha that's totally fair - I think if we go down the route of only accepting the HMAC algorithms, we could define that subset as a constant and make that the input to this function instead.

oidc/clientassertion/client_assertion.go
johanbrandhorst: I think we probably should prevent it, so that it can only be set with the `WithKeyID` option. Does that make sense?

oidc/clientassertion/doc.go
```diff
@@ -0,0 +1,13 @@
+package clientassertion
```
mgaffney, line 1: What about a package doc string?

gulducat: good point, I'll do this tomorrow when I make an `example_test.go`

gulducat:
@johanbrandhorst I believe I've addressed the bulk of your feedback. Let me know what ya think?

I'll fix the `go.mod` mergeflict tomorrow (if I keep going tonight, I'll get carried away :yum:), and also an `example_test.go` along with a package docstring.

gulducat:
Alrighty folks, ready for another pass!

oidc/clientassertion/client_assertion.go
```diff
@@ -0,0 +1,201 @@
...
+		return nil, errors.Join(errs...)
+	}
+
+	if err := j.Validate(); err != nil {
+		return nil, fmt.Errorf("new client assertion validation error: %w", err)
+	}
+	return j, nil
+}
+
+// JWT signs a JWT with either a private key or a secret
```
jimlambrt, line 65:
```suggestion
// JWT is used to create a client assertion JWT which is a special JWT used by an OAuth 2.0 or OIDC client to authenticate themselves to an authorization server.
```

oidc/clientassertion/client_assertion.go
```diff
@@ -0,0 +1,201 @@
...
+	// finally, make sure Serialize() works; we can't pre-validate everything,
+	// and this whole thing is useless if it can't Serialize()
+	if _, err := j.Serialize(); err != nil {
+		return fmt.Errorf("serialization error during validate: %w", err)
+	}
+
+	return nil
+}
+
+// Serialize returns a signed JWT string
```
jimlambrt, line 127:
```suggestion
// Serialize returns client assertion JWT which can be used a an OAuth 2.0 or OIDC client to authenticate themselves to an authorization server
```

oidc/provider.go
```diff
@@ -306,6 +307,18 @@ func (p *Provider) Exchange(ctx context.Context, oidcRequest Request, authorizat
 	if oidcRequest.PKCEVerifier() != nil {
 		authCodeOpts = append(authCodeOpts, oauth2.SetAuthURLParam("code_verifier", oidcRequest.PKCEVerifier().Verifier()))
 	}
+	if oidcRequest.ClientAssertionJWT() != nil {
+		// by now, sufficient validation should have already occurred to prevent
+		// errors here, but err check again just in case.
+		token, err := oidcRequest.ClientAssertionJWT().Serialize()
+		if err != nil {
+			return nil, err
```
jimlambrt, line 315: Perhaps follow the existing pattern and format for returned errors (incl. the `op`)

jimlambrt (changes requested):
Ty for the thoughtful contribution. Just a few minor suggestions.

oidc/clientassertion/client_assertion.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+// Serializer is the primary interface impelmented by JWT.
+type Serializer interface {
+	Serialize() (string, error)
+}
+
+// ensure JWT implements Serializer, which is accepted by the oidc option
+// oidc.WithClientAssertionJWT.
+var _ Serializer = &JWT{}
```
johanbrandhorst, line 201: I don't think we really need this interface, or at the very least we don't need to export it. In practice we should notice very quickly if `JWT` is no longer usable with `oid.JWTSerializer`.

oidc/clientassertion/options.go
```diff
@@ -0,0 +1,70 @@
...
+// * HS384 with a >= 48 byte secret
+// * HS512 with a >= 64 byte secret
+func WithClientSecret(secret string, alg HSAlgorithm) Option {
+	return func(j *JWT) error {
+		if err := alg.Validate(secret); err != nil {
+			return err
+		}
+		j.secret = secret
+		j.alg = jose.SignatureAlgorithm(alg)
+		return nil
```
johanbrandhorst, line 28: Maybe this should error if `j.key` is set?

oidc/clientassertion/options.go
```diff
@@ -0,0 +1,70 @@
...
+// * RS256
+// * RS384
+// * RS512
+func WithRSAKey(key *rsa.PrivateKey, alg RSAlgorithm) Option {
+	return func(j *JWT) error {
+		if err := alg.Validate(key); err != nil {
+			return err
+		}
+		j.key = key
+		j.alg = jose.SignatureAlgorithm(alg)
```
johanbrandhorst, line 43: This could error if j.secret is set.

oidc/clientassertion/options.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+}
+
+// WithHeaders sets extra JWT headers
+func WithHeaders(h map[string]string) Option {
+	return func(j *JWT) error {
+		for k, v := range h {
+			// disallow potential confusion arising from the "kid" header
+			// being set both by this and WithKeyID()
+			if k == "kid" {
```
johanbrandhorst, line 63: Nit: maybe make the `"kid"` value a constant with a reference to the definition of this header?

johanbrandhorst (changes requested):
Thanks for the update, it made me think of some new feedback, hope it makes sense!

oidc/clientassertion/algorithms.go
```diff
@@ -0,0 +1,55 @@
...
+	case HS256:
+		expectLen = 32
+	case HS384:
+		expectLen = 48
+	case HS512:
+		expectLen = 64
+	default:
+		return fmt.Errorf("%w %q for client secret", ErrUnsupportedAlgorithm, a)
+	}
+	if len(secret) < expectLen {
```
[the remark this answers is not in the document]
gulducat: From what I have gathered, the secret can be longer! it just needs to be _at least_ yay long.  Keycloak, for example, provided me with a `33` character secret.

This seems to be the definition: https://datatracker.ietf.org/doc/html/rfc2104#section-3

> Keys longer than L bytes are acceptable but the extra length would not significantly increase the function strength.

oidc/clientassertion/client_assertion.go
```diff
@@ -0,0 +1,201 @@
...
+type JWT struct {
+	// for JWT claims
+	clientID string
+	audience []string
+	headers  map[string]string
+
+	// for signer
+	alg jose.SignatureAlgorithm
+	// key may be any key type that jose.SigningKey accepts for its Key
+	key any
```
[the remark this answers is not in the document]
gulducat: Yeah currently the only `With*Key` option is RSA, but I figured I'd leave this open for someone to extend later. Think I should close it off, and let hypothetical Later Person figure out that they can change this to `any`?

oidc/clientassertion/client_assertion.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+
+	// finally, make sure Serialize() works; we can't pre-validate everything,
+	// and this whole thing is useless if it can't Serialize()
+	if _, err := j.Serialize(); err != nil {
+		return fmt.Errorf("serialization error during validate: %w", err)
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
gulducat: Seems reasonable enough to me -- Okay if I keep it as a separate function, just privatize it? I like small single-purpose functions.

oidc/clientassertion/options.go
gulducat: I was letting `validate()` handle that, since it's a cross-attribute check, and it's already ensuring that at least one of them is set.

gulducat:
Thanks y'all! I think I've addressed all your comments, with a couple exceptions that I could be convinced of. :)  Let me know what ya think!

jimlambrt (changes requested):
Ty for all the changes. I had just 2 more I hoping you'll consider.

oidc/clientassertion/algorithms.go
```diff
@@ -0,0 +1,74 @@
...
+
+// Validate checks that the secret is a supported algorithm and that it's
+// the proper length for the HSAlgorithm:
+//   - HS256: >= 32 bytes
+//   - HS384: >= 48 bytes
+//   - HS512: >= 64 bytes
+func (a HSAlgorithm) Validate(secret string) error {
+	const op = "HSAlgorithm.Validate"
+	if secret == "" {
+		return fmt.Errorf("%w: empty", ErrInvalidSecretLength)
```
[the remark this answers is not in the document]
gulducat: shoot, missed this one!

oidc/provider.go
gulducat: apologies, I just missed it! my eyes might have started to glaze over after doing a couple dozen of em :stuck_out_tongue_closed_eyes:

oidc/clientassertion/algorithms.go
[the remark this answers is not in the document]
gulducat: fixed! and the other you found too

jimlambrt (approved):
Ty again for the contribution.

gulducat:
@johanbrandhorst let me know if you'd like to press on any of your unresolved comments, or if there's anything else I can do. :)

oidc/clientassertion/client_assertion.go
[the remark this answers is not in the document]
johanbrandhorst: Sure, but I think some of this error checking doesn't really belong here anymore. E.g. the checking if j.key and j.secret are both set aren't necessary if you move the logic into the option as I suggested. Calling `Serialize` in `New` also feel really wrong IMO, since it's the primary thing the user wants to do. If that's where we're going with this - why do we need a whole struct (or even package) encapsulating this? We could just have `NewSerializedJWT` as an exported function (maybe we should?). I think `New()` and `Serialize()` can be separate, but if this logic was moved into `New`, it would be clear from the context of that function what should and shouldn't be necessary to validate. Ideally nothing other than the inputs to the function, which should be all the required parameters to create a JWT, should need to be validated. On that note, since either Key or Secret _must_ be set, should it be a parameter rather than an option?

I'm personally not a huge fan of more smaller functions, because the context in which they are used is not clear and indirection always leads to overhead for the reader. I almost left a comment saying that the structure in `Serialize` was unnecessarily abstracted into smaller functions, but it's such a minor comment that I didn't in the end feel that it was important enough to bring up, but since we're talking about it now, there you go 😁.

oidc/clientassertion/client_assertion.go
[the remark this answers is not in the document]
johanbrandhorst: I'd like to restrict this to the types we support right now, if that's okay.

oidc/clientassertion/algorithms.go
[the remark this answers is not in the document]
johanbrandhorst: That makes sense, thanks! Mind adding a comment in the code here explaining this?

oidc/clientassertion/algorithms.go
```diff
@@ -0,0 +1,74 @@
...
+type (
+	// HSAlgorithm is an HMAC signature algorithm
+	HSAlgorithm string
+	// RSAlgorithm is an RSA signature algorithm
+	RSAlgorithm string
+)
+
+const (
+	// JOSE asymmetric signing algorithm values as defined by RFC 7518.
+	// See: https://tools.ietf.org/html/rfc7518#section-3.1
```
johanbrandhorst, line 20: If you put this comment above the `const` block it will apply to all the consts, I think.

oidc/clientassertion/client_assertion.go
```diff
@@ -0,0 +1,202 @@
...
+		Subject:   j.clientID,
+		Audience:  j.audience,
+		Expiry:    jwt.NewNumericDate(now.Add(5 * time.Minute)),
+		NotBefore: jwt.NewNumericDate(now.Add(-1 * time.Second)),
+		IssuedAt:  jwt.NewNumericDate(now),
+		ID:        id,
+	}
+}
+
+// serializer is the primary interface impelmented by JWT.
```
johanbrandhorst, line 195:
```suggestion
// serializer is the primary interface implemented by JWT.
```

oidc/clientassertion/options.go
```diff
@@ -0,0 +1,79 @@
...
+
+// WithKeyID sets the "kid" header that OIDC providers use to look up the
+// public key to check the signed JWT
+func WithKeyID(keyID string) Option {
+	const op = "WithKeyID"
+	return func(j *JWT) error {
+		if keyID == "" {
+			return fmt.Errorf("%s: %w", op, ErrMissingKeyID)
+		}
+		j.headers["kid"] = keyID
```
johanbrandhorst, line 61:
```suggestion
		j.headers[KeyIDHeader] = keyID
```

johanbrandhorst:
@gulducat looks like there are still some unresolved replies on the existing comments that I replied to

gulducat:
> still some unresolved replies

My bad, I got ahead of myself seeing only the new ones. :heart:

oidc/clientassertion/client_assertion.go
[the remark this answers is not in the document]
gulducat: Thanks for the thoughtful reply!

I'll say up front that I'm down to remove the `Serialize()` call from `New` - it felt weird when I put it there, so easy to drop it.  Also good to move validation into `New`.

> the checking if j.key and j.secret are both set aren't necessary if you move the logic into the option

I like cross-field validation to happen in its own spot, but it amounts to the same result either way, so I'd be happy to put the checks in their respective options. except:

> since either Key or Secret must be set, should it be a parameter rather than an option?

The current not-actually-optional-`Option` thing has been bugging me.  Here's some (pardon the pun) other options:

1. I don't love funcs that take multiple params where any of them are expected to always be zero values like `""` / `nil`
1. We could do similar to what `jose` does, and accept an `any` but assert supported types. Personally this kind of annoys me about `jose`
1. Or, what if we have 2 constructors?

```go
func NewJWTWithRSAKey(clientID string, audience []string, key *rsa.RSAKey, opts ...Option) (*JWT, error) {
}
func NewJWTWithHMAC(clientID string, audience []string, secret string, opts ...Option) (*JWT, error) {
}
```

That would leave a lot less validation to do, but both could set up and return a `JWT`, which is mostly shared logic internally.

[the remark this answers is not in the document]
johanbrandhorst: I like option 3 a lot!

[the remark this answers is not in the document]
johanbrandhorst: I believe the algorithm will need to be a parameter too, and you could even make it RSAlgorithm and HSAlgorithm respectively :)

gulducat:
Got a big ol refactor for ya @johanbrandhorst, spurred by https://github.com/hashicorp/cap/pull/155#discussion_r1972188537. I do think it's a nice improvement to the ergonomics.

I think I covered your other smaller concerns, too?

The diff is pretty large (including completely rewritten tests), but the public interface is only slightly so, as demonstrated by the barely-changed [`example_test.go`](https://github.com/hashicorp/cap/pull/155/commits/e8dfe1983ebefb36957b082f01a62950e1d9e4c5#diff-aea07b3a6c30d3b78eaa9bc69fa29dc9b186c4887b4d801f07e51f285b02bf8e)

johanbrandhorst (dismissed):
This looks great, thank you so much for being so responsive and accepting of feedback! I just had a few nits left but I'm happy to merge as is if you want.

oidc/clientassertion/client_assertion_test.go
```diff
@@ -0,0 +1,208 @@
...
+	err := errors.Unwrap(actual)
+	joined, ok := err.(joinedErrs)
+	require.True(t, ok, "expected Join()ed errors; got: %v", actual)
+	unwrapped := joined.Unwrap()
+	require.ElementsMatch(t, expect, unwrapped)
+}
+
+// TestJWTBare tests what errors we expect if &JWT{}
+// is instantiated directly, rather than using a constructor.
+func TestJWTBare(t *testing.T) {
```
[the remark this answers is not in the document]
gulducat: We can, yep; they're well isolated. I did some test duration math though, and this is how adding parallels shook out:

* none: 0.2998s
* top-level: 0.2414s
* & subtests: 0.2311s

Each time is an average of 10 runs on my pretty nice linux laptop. It seems pretty small savings to me, since this is a pretty lightweight package.

I kept `t.Parallel()` in the top-level tests like this one, but IMO it's more visual noise than its worth in all the (now separate and quite short) subtests.

gulducat:
Thanks for helping me make this thing the best it can be!  Latest feedback commit is both easier and harder to read ignoring whitespace: https://github.com/hashicorp/cap/pull/155/commits/80514102ada6c41df3e50bd02e12212c9e496c14?w=1 -- Conveniently, I could just turn my code comments into subtest names. :)

johanbrandhorst (approved):

PR #159, opened 2025-03-20
Vault 33101 ignore trailing slash

jaireddjawed:


jimlambrt (commented):
Can you update the PR description to something that doesn't describe the change related to a specific vault plugin?

Basically, you should give some background information related to audiences (and likely the issuer as well - see below) and this change is being proposed and the implications of making the change.  This should include references to the appropriate internet RFCs about this very topic and in particular URI comparisons.  This will allow the PR reviewer, future maintainers, and pkg consumers to understand why this change was made and the implications to every pkg consumer.    There were several slack threads that might help inform this more complete description.

jimlambrt (commented):
Can you use the existing options pattern that's used throughout pkgs in the repo?

jimlambrt (changes requested):
Can you add the appropriate unit tests to this PR which cover all proposed changes?

jaireddjawed:
> Basically, you should give some background information related to audiences (and likely the issuer as well - see below) and this change is being proposed and the implications of making the change. This should include references to the appropriate internet RFCs about this very topic and in particular URI comparisons. This will allow the PR reviewer, future maintainers, and pkg consumers to understand why this change was made and the implications to every pkg consumer. There were several slack threads that might help inform this more complete description.

Sorry about that. Was in the middle of updating the description when you were reviewing.

jimlambrt (changes requested):
Please make all the changes requested before asking for re-review. BTW, you also need a changelog entry.

Also the updated description doesn't quite see the change I was requesting.

jaireddjawed:
> Can you add the appropriate unit tests to this PR which cover all proposed changes?

Unit tests were already added.

jaireddjawed:
> Can you update the PR description to something that doesn't describe the change related to a specific vault plugin?
> 
> This should include references to the appropriate internet RFCs about this very topic and in particular URI comparisons. 

I don't believe that there was an RFC doc for this change.

PR #161, opened 2025-06-04
dep: Pin action/cache to a contianer sha

louisruch:


louisruch:
@johanbrandhorst I just bumped this to 4.2.3 at the same time

johanbrandhorst (approved):

PR #172, opened 2025-11-28
Expose OIDC provider server metadata in Provider

allisonlarson:
This exposes the OIDC Provider auth server metadata that is returned during service discovery, and exposed by go-oidc, and adds a way to supply it during testing. This allows OP metadata to be accessed on the provider object once its  been discovered, instead of making an additional request to the same endpoint, or having to have explicit support.

Hi! I'm an engineer on Nomad and I'm hoping to add some support for RFC 9207 (OAuth 2.0 Authorization Server Issuer Identification), which involves two steps during the OIDC Complete Authorization step of the login flow:
* Check the value of `authorization_response_iss_parameter_supported` of the OP auth server metadata (returned in the `.well-known/openid-configuration` endpoint)
* Require the `iss` param value in the Auth Response is present & matches based on OP support

Currently there isn't a way to check the auth server metadata via the cap/oidc library for an unsupported configuration value. Once we have a way to access the metadata, we can determine if we should manually check the `iss` param in the auth response. The `authorization_response_iss_parameter_supported` isn't required to be set, so the approach I took was to supply a way to get the configuration value out; I'm happy to discuss any other thoughts about the best way to approach. I'm hoping this will help any other users of the library support the same functionality.

The approach is to expose the same `Claims()` function on the [Provider as defined in go-oidc](https://github.com/coreos/go-oidc/blob/v3/oidc/oidc.go#L304), as a new function with a unit test.

The test provider has been updated to be able to supply additional server metadata that gets merged and returned in the `openidConfiguration` path. (The test was implemented with an arbitrary config value, but I can update it to something realistic)

## PCI review checklist

- [x] I have documented a clear reason for, and description of, the change I am making.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ Canonical reference for changes, improvements, and bugfixes for cap.
 
 ## Next
 
+* feat (oidc): add Claims for exposing provider server metadata ([PR #172](https://github.com/hashicorp/cap/pull/172))
+
 ## 0.11.0
 
 * fix (ldap): fix slice append to be concurrent safe when searching for ldap


jimlambrt (approved):
Looks good! Ty.

jimlambrt (approved):
Ty

PR #186, opened 2026-04-09
JWT: add optional SignatureVerifier callback to Validator

suraj-simha:
Adds an optional SignatureVerifier callback to the JWT Validator to support custom signature verification logic. When provided, the callback is called instead of the default sequential KeySet verification. 

Motivation:
Vault's JWT auth plugin uses CAP for JWT validation, which delegates signature verification to `go-oidc` . The current flow checks each keyset's cache and if the key is not found, fetches from the JWKS URL before moving to the next keyset. This sequential fetch-per-keyset behavior was causing a performance issue for a customer with multiple JWKS endpoints configured.

This change adds a SignatureVerifier callback to CAP's Validator, allowing the plugin to own the signature verification logic and implement the desired behavior of checking all cache's before refreshing any key from the remote JWKs URLs.

The `NewValidatorWithSignatureVerifier` has been added to avoid modifying the existing `NewValidator` signature. Existing callers using `NewValidator` are completely unaffected.

## PCI review checklist

- [ ] I have documented a clear reason for, and description of, the change I am making.


santoshhashicorp (approved):
LGTM

PR #187, opened 2026-04-15
JWT: add optional KeySetSearcher callback to Validator

suraj-simha:
Adds an optional `KeySetSearcher` callback to the JWT Validator to support dynamic KeySet lookup based on the JWT's key ID (kid) header. When provided, the Validator extracts the kid from the JWT header and calls the callback to retrieve the appropriate KeySet, instead of the default sequential KeySet verification.

Motivation:
Vault's JWT auth plugin uses CAP for JWT validation, which delegates signature verification to `go-oidc`. The current flow checks each keyset's cache and if the key is not found, fetches from the JWKS URL before moving to the next keyset. This sequential fetch-per-keyset behavior was causing a performance issue for a customer with multiple JWKS endpoints configured.

This change adds a `KeySetSearcher` callback to CAP's Validator, allowing the plugin to implement custom KeySet lookup logic. The Validator handles all JWT parsing and security validation (extracting the kid from the JWT header, verifying the signature with the returned KeySet), while the callback provides the logic for locating the appropriate KeySet. This enables implementations to check all caches before refreshing any keys from remote JWKS URLs.

The `NewValidatorWithKeySetSearcher` constructor has been added to avoid modifying the existing `NewValidator`. Existing callers using `NewValidator` are completely unaffected.

## PCI review checklist

- [ ] I have documented a clear reason for, and description of, the change I am making.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ Canonical reference for changes, improvements, and bugfixes for cap.
 ## Next
 * fix (saml): always validate response and assertion signatures when signatures are present ([PR #180](https://github.com/hashicorp/cap/pull/180))
 
+* feat (jwt): add optional KeySetSearcher callback to Validator for dynamic KeySet lookup ([PR #187](https://github.com/hashicorp/cap/pull/187))
+
 ## 0.12.0
 * feat (oidc): add Claims for exposing provider server metadata ([PR #172](https://github.com/hashicorp/cap/pull/172))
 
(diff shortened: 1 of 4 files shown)


jimlambrt (changes requested):
Do we want explicit tests for error cases for things like multi-signatures, missing kids, malformed token?

suraj-simha:
> Do we want explicit tests for error cases for things like multi-signatures, missing kids, malformed token?

Yep, added tests for those cases. Thanks for calling it out!

jimlambrt (approved):
Ty

PR #193, opened 2026-08-27
feat(oidc): Creation Of Azure Package For Resolving Group Overage Claims

RyanDerr:
# Overview
This PR adds a new `oidc/azure` package that aims to resolve Azure Entra ID distributed group claims. When a user belongs to more than 200 groups, Azure omits the groups claim from the token and replaces it with an [overage indicator](https://learn.microsoft.com/en-us/entra/identity-platform/access-token-claims-reference#groups-overage-claim) pointing to the Microsoft Graph API. This package exposes logic to detect when an overage is present in a token's claims, and fetch and return the full list of groups associated with the user when present. 

# Changes
- Creation of a `azure` package which contains new function(s) `ResolveGroupClaims` and helper functions for detecting the overage indicator, resolving the correct regional Graph API endpoint, and fetching group IDs. 
   - This package supports various Azure cloud deployment endpoints (both [global and national](https://learn.microsoft.com/en-us/graph/deployments#microsoft-graph-and-graph-explorer-service-root-endpoints)) and resolves [deprecated AAD hosts](https://learn.microsoft.com/en-us/graph/migrate-azure-ad-graph-request-differences) automatically.   
- Addition of sentinel errors within the package to support input checks. 
- Package level doc explaining the overage scenario in Azure for groups, and providing official MSFT docs for when it's trigger and how to resolve them.

### DIFF
--- a/oidc/azure/azure.go
+++ b/oidc/azure/azure.go
@@ -1,0 +1,238 @@
+// Copyright IBM Corp. 2026
+// SPDX-License-Identifier: MPL-2.0
+
+package azure
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"golang.org/x/oauth2"
+)
+
+// azureGroups is the structured response from the Microsoft Graph API when fetching group IDs for a user.
+// The Value field contains a slice of group IDs.
+type azureGroups struct {
+	Value []string `json:"value"`
+}
+
+const (
+
+	// Deprecated: The Azure AD (AAD) Graph API hosts.
+	azureADGraphHost   = "graph.windows.net"
+	azureADGraphUSHost = "graph.microsoftazure.us"
+	azureADGraphCNHost = "graph.chinacloudapi.cn"
+
+	// The Microsoft Graph API hosts.
+	microsoftGraphHost      = "graph.microsoft.com"
+	microsoftGraphUSHost    = "graph.microsoft.us"
+	microsoftGraphUSDodHost = "dod-graph.microsoft.us"
+	microsoftGraphCNHost    = "microsoftgraph.chinacloudapi.cn"
+
+	// getMemberObjectsPath is the path to the Microsoft Graph API endpoint for
+	// retrieving the groups a user is a member of.
+	getMemberObjectsPath = "/v1.0/me/getMemberObjects"
+
+	// claimNameField and claimsSourceField are the Azure distributed claims fields.
+	claimNameField   = "_claim_names"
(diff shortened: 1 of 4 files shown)


oidc/azure/azure.go
```diff
@@ -0,0 +1,235 @@
...
+	microsoftGraphUSDodHost: microsoftGraphUSDodHost,
+	microsoftGraphCNHost:    microsoftGraphCNHost,
+}
+
+// ResolveGroupClaims will detect an Azure groups overage indicator in the
+// provided claims and fetch the full list of group IDs from the Microsoft
+// Graph API using the provided OAuth2 token. When no overage indicator is
+// present, it returns an empty map. On success, the returned map contains
+// a "groups" key populated with the user's group IDs.
+func ResolveGroupClaims(ctx context.Context, client *http.Client, token *oauth2.Token, claims map[string]any) (map[string]any, error) {
```
RyanDerr, line 72: **Open Question**: Depending on how we want to treat the parent oidc lib to be more azure specific, wondering if should we return a map or just a slice of the group ids to persist and merge in the future with the claims from the resulting id token?

oidc/azure/azure.go
```diff
@@ -0,0 +1,235 @@
+// Copyright IBM Corp. 2020, 2026
```
louisruch: shouldn't this just be 2026

oidc/azure/azure.go
```diff
@@ -0,0 +1,235 @@
...
+//	    },
+//	    "_claim_sources": {
+//	        "src1": {
+//	            "endpoint": "https://graph.microsoft.com/v1.0/users/{userID}/getMemberObjects"
+//	        }
+//	    }
+//	}
+func fetchGroupsOverageHost(claims map[string]any) (string, bool) {
+	// Check for the presence of the _claim_names field and ensure it's a map.
+	claimNames, ok := claims[claimNameField].(map[string]any)
```
louisruch, line 134: Did we wanna nil check claims before

louisruch (commented):
Two nits but otherwise this looks great

oidc/azure/azure.go
RyanDerr: Yeah I think you're right. I mainly just copy pastaed this following the same Copyright structure elsewhere but I it likely just needs to be `// Copyright IBM Corp. 2026`.

oidc/azure/azure.go
RyanDerr: Yep we should, good callout!

louisruch (approved):

bgajjala8 (approved):

PR #194, opened 2026-08-27
feat(oidc): Creation Of New Provider Type Option For Provider Config

RyanDerr:
# Overview
This PR adds a new optional `ProviderType` field to the OIDC Config struct. The purpose of this option is to allow callers to identify which OIDC provider they are configuring so that provider specific logic can be triggered in cap. 

Currently the only initial supported type is `Azure`, which will enable in the future resolving group overage claims.

# Changes
- Creation of a new file `provider_type.go` within the oidc package which contains the following: 
   -  `ProviderType` string type
   -  `ProviderTypeAzure` constant to denote the `Azure` provider
   - A `SupportedProviderType` helper function that validates if a given provider type is supported within the library
- Addition of the `ProviderType` field to the `Config` struct in addition to the `WithProviderType` option function that sets `ProviderType` on a `Config` during construction.
- Modification of the `Config` methods: 
   - `Validate` which now rejects and throws an error any non-empty provider type `ProviderType` supplied that is not currently known and supported. 
   - `Hash` adds the `ProviderType` in the hash input to be computed, such that the hash is consistent across all config fields.

### DIFF
--- a/oidc/provider_type.go
+++ b/oidc/provider_type.go
@@ -1,0 +1,22 @@
+// Copyright IBM Corp. 2026
+// SPDX-License-Identifier: MPL-2.0
+
+package oidc
+
+// ProviderType identifies a given known OIDC provider.
+type ProviderType string
+
+const (
+	// ProviderTypeAzure is the provider identifier for Microsoft Azure.
+	ProviderTypeAzure ProviderType = "azure"
+)
+
+// supportedProviderTypes is a map of known OIDC provider types.
+var supportedProviderTypes = map[ProviderType]bool{
+	ProviderTypeAzure: true,
+}
+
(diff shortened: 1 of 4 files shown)


oidc/provider_type.go
```diff
@@ -0,0 +1,22 @@
+// Copyright IBM Corp. 2020, 2026
```
louisruch: Same question on if we need 2020 here

louisruch (approved):
LGTM

oidc/provider_type.go
RyanDerr: Yee

bgajjala8 (approved):

PR #198, opened 2026-09-09
feat(oidc): resolve Azure group claims during token exchange

bgajjala8:
# Overview
This PR integrates the Azure group overage claim resolution into the OIDC authorization code exchange.

When the provider is configured with `ProviderTypeAzure`, `Provider.Exchange` now resolves distributed group claims through the existing Azure package and stores the resolved claims on the returned token.

# Changes
- Update `Provider.Exchange` to call `azure.ResolveGroupClaims` for Azure providers.
- Store the resolved Azure claims in the returned token’s `additionalClaims`.
- Update test to cover azure provider specific call.


oidc/provider_test.go
```diff
@@ -705,6 +705,101 @@ func TestProvider_Exchange(t *testing.T) {
...
+	if req.URL.Hostname() == "graph.microsoft.com" {
+		return &http.Response{
+			StatusCode: http.StatusOK,
+			Body: io.NopCloser(
+				strings.NewReader(`{"value":["group-1","group-2"]}`),
+			),
+			Header:  make(http.Header),
+			Request: req,
+		}, nil
+	}
```
bgajjala8: I know we talked about exporting the Azure test function for the `RoundTripper`. I ended up keeping it in the `oidc` package because I wasn’t sure if we want to expose a test only helper by exporting it. If ya'll have any other thoughts or think there is a better approach, I’d love to hear them. I chose not to DRY it up because right now it only applies to Azure, and we don’t have a clearly defined pattern yet.

oidc/provider_test.go
```diff
@@ -705,6 +705,101 @@ func TestProvider_Exchange(t *testing.T) {
 	})
 }
 
+func TestProvider_Exchange_Azure(t *testing.T) {
```
RyanDerr: **Question**: Is there a reason to separate this test out from `TestProvider_Exchange`? Was there any setup that was unique to this test that cannot be done within a new `t.Run`, as the `TestProvider_Exchange` test function does not use test tables.

oidc/provider.go
```diff
@@ -366,6 +367,19 @@ func (p *Provider) Exchange(ctx context.Context, oidcRequest Request, authorizat
...
+	if p.config.ProviderType == ProviderTypeAzure {
+		client, err := p.HTTPClient()
+		if err != nil {
+			return nil, fmt.Errorf("%s: unable to create Azure HTTP client: %w", op, err)
+		}
+		claims, err := azure.ResolveGroupClaims(ctx, client, oauth2Token, claims)
+		if err != nil {
+			return nil, fmt.Errorf("%s: unable to resolve Azure group claims: %w", op, err)
+		}
+		t.additionalClaims = claims
```
RyanDerr, line 382: **Suggestion**: I think we should leave a `TODO` comment that eventually as the need for additional claims or providers expands, we should merge the fetched claims with any already set additional claims. Currently this is okay now since there's only ever one call site, but as this expands we likely don't want to always overwrite the map.

oidc/provider_test.go
RyanDerr: I think for right now it is okay to expose a helper function from the `azure` package to mock the HTTP request and response pattern we expect from Azure. My main concern is avoiding drift between the `azure` package and what it's validating and testing and what the caller is using to mock behaviors. If they mock Azure differently, we could end up with misleading or flaky tests. We may need to refactor the `azure` package or create a new test function to give back mocked response data to support callers tests, but I think centralizing this behavior makes the most sense for the moment.

oidc/provider_test.go
bgajjala8: I kept it separate primarily to preserve readability and isolate the Azure specific setup, although it can also be done inside the existing `TestProvider_Exchange` function. I’m happy to move it there if that better aligns with the testing conventions.

RyanDerr: For now I think it's okay to place them in the primary exchange test and have an explicit `t.Run(...)` name for the subtest.

bgajjala8: Gotcha updated here - https://github.com/hashicorp/cap/pull/198/changes/f0d95bc85fa20e81248ceeca866e5db00849680e

oidc/provider_test.go
bgajjala8: Gotcha, I updated it here - [145f2db](https://github.com/hashicorp/cap/pull/198/commits/145f2db5cbed5ec24cba9766fd890f70bdba1e6e). However, since this is a widely used library, could exposing the test helper to consumers have any unintended implications? I’m not sure whether this is a concern, or if there is a better way to expose this.

RyanDerr: I just realized that we may need to rethink this strategy, as this exposes the helpers outside of just this package itself.

irenarindos (approved):

RyanDerr (approved):
Nice job!

PR #200, opened 2026-09-11
Azure EntraID Group Overage Claim Resolution 

RyanDerr:
## Overview
Azure Entra ID returns distributed group claims when a user belongs to more than 200 groups. The purpose of this PR is to update CAP to resolve those claims through the Microsoft Graph API during OIDC token exchange, so downstream authorization receives the user's complete group membership(s).

## Changes
- Creation of new `azure` subpackage within `oidc` to detect and resolve Azure Entra group IDs from OIDC token overage claims through Microsoft Graph. API. 
- Updates to the `Tk` struct to stored resolved group claims and for a new `ProviderType` to support for provider specific overage claim resolution while preserving existing auth exchange behavior by default. 
- Addition of the `EffectiveClaims` method on a `Tk` to merge exchanged and resolved claims with OIDC token claims taking precedence.


oidc/provider.go
```diff
@@ -366,6 +367,21 @@ func (p *Provider) Exchange(ctx context.Context, oidcRequest Request, authorizat
...
+	if p.config.ProviderType == ProviderTypeAzure {
+		client, err := p.HTTPClient()
+		if err != nil {
+			return nil, fmt.Errorf("%s: unable to create Azure HTTP client: %w", op, err)
+		}
+		claims, err := azure.ResolveGroupClaims(ctx, client, oauth2Token, claims)
+		if err != nil {
+			return nil, fmt.Errorf("%s: unable to resolve Azure group claims: %w", op, err)
+		}
+		// TODO: Merge fetched claims with existing additional claims when
```
irenarindos, line 380: Just out of curiosity is this something planned in the future?

oidc/azure/azure.go
```diff
@@ -0,0 +1,238 @@
...
+		return "", false
+	}
+
+	// Parse the endpoint URL to extract the host and validate it against known Graph API hosts.
+	u, err := url.Parse(endpoint)
+	if err != nil {
+		return "", false
+	}
+
+	if !knownGraphHost(u.Host) {
```
irenarindos: Should this use u.Hostname() to strip any ports?

oidc/provider_type.go
```diff
@@ -0,0 +1,22 @@
...
+type ProviderType string
+
+const (
+	// ProviderTypeAzure is the provider identifier for Microsoft Azure.
+	ProviderTypeAzure ProviderType = "azure"
+)
+
+// supportedProviderTypes is a map of known OIDC provider types.
+var supportedProviderTypes = map[ProviderType]bool{
+	ProviderTypeAzure: true,
```
irenarindos: Pontentially a nit.. would we add a provider type that isn't supported? Could we use map[ProviderType]struct{} with an empty struct instead?

oidc/provider.go
RyanDerr: This isn’t planned for the immediate future. However, if another provider requires resolving additional claims from the OIDC token during the exchange, or if Azure requires additional fetching, we’ll need to update this logic. This is mainly a note for the future in case we go that route.

oidc/azure/azure.go
RyanDerr: That's a great callout, yes we should. This was mainly an oversight since our testing was using DNS names without a port, but we in reality should. I've updated this logic and added a testcase for this path.

oidc/provider_type.go
RyanDerr: I think that's fair, we don't necessarily need it to be a `map[ProviderType]bool` as all future `ProviderType` added should be supported. In reality we just need to know if they're recognized for support or not.

irenarindos (approved):
Thanks for this!

